PR #18, opened 2018-09-11
Wip/meta language - class syntax

rayslxu:


lib/class_def_ast.js
```diff
@@ -0,0 +1,340 @@
...
+            metadata, annotations);
+    }
+
+    toString() {
+        let annotations = '';
+        Object.entries(this.metadata).forEach(([name, value]) => {
+            annotations += ` #_[${name}=${prettyprintValue(value)}]`;
+        });
+        Object.entries(this.annotations).forEach(([name, value]) => {
+            annotations += ` #_[${name}=${prettyprintValue(value)}]`;
```
gcampax: This should just be `#[]`
(we might need better codegen tests...)

lib/class_def_ast.js
```diff
@@ -0,0 +1,340 @@
...
+        return this._annotations;
+    }
+
+    toString(prefix = '') {
+        let annotations = '';
+        Object.entries(this.metadata).forEach(([name, value]) => {
+            annotations += `\n${prefix}#_[${name}=${prettyprintValue(value)}]`;
+        });
+        Object.entries(this.annotations).forEach(([name, value]) => {
+            annotations += `\n${prefix}#_[${name}=${prettyprintValue(value)}]`;
```
gcampax: Same here

lib/grammar.pegjs
```diff
@@ -503,21 +538,44 @@ array_value_list = first:value _ rest:(',' _ value _)* {
...
+    return { type: 'nl', name: name, value: value };
+}
+impl_annotation = '#' '[' name:ident _ '=' value:value ']' {
+    return { type: 'impl', name: name, value: value };
+}
+
 // Types
 
-type_ref = 'Measure' _ '(' _ unit:ident? _ ')' { return Type.Measure(unit); } /
+type_ref = hashable_type_ref /
```
gcampax: do you still use this distinction?

lib/utils.js
```diff
@@ -86,6 +86,10 @@ module.exports = {
             return schemaRetriever.getSchemaAndNames(type, schemaType, name).then((schema) => module.exports.splitArgsForSchema(schema, schemaType, false));
     },
 
+    getMixin(schemaRetriever, kind) {
+        return schemaRetriever.getMixins(kind);
+    },
+
```
gcampax: do you need this indirection?

lib/grammar.pegjs
gcampax: if we don't need it, it's just extra cognitive load on our part when we read this code later on; as in, we'll come back 3 months from now to this code and wonder: "what was this again?" "is my new type hashable or not?"

lib/utils.js
rayslxu: just trying to be consistent with getschema, is it bad?

lib/class_def_ast.js
rayslxu: fixed

lib/class_def_ast.js
rayslxu: fixed

lib/utils.js
gcampax: It's extra code that does not add much to readability. So yeah I'd get rid of it.

PR #19, opened 2018-09-14
Add dataset language to ThingTalk

rayslxu:


gcampax:
What happened that made us lose coverage? Did we drop tests?

lib/grammar.pegjs
```diff
@@ -145,6 +149,26 @@ param_direction = 'in' __ 'req' { return 'in req'; } / 'in' __ 'opt' { return 'i
     'out' { return 'out'; }
 
 
+// Dataset & Thingpedia Definition
+dataset_name = ident / name:class_name { return '@' + name; }
+
+dataset = _ 'dataset' _ name:dataset_name _ lan:('language' _ literal_string _)? '{' _ examples:(_ example _)* _ '}' {
+    return new Ast.Dataset(name, lan? lan[2] : 'en' , take(examples, 1));
+}
```
gcampax: Should we give structured metadata to datasets? Most other stuff has it, so for consistency I feel like we should.

Also, classes probably should have metadata (which we overlooked in the class merge) to cover User Visible Name and User Visible Description.

lib/grammar.pegjs
```diff
@@ -145,6 +149,26 @@ param_direction = 'in' __ 'req' { return 'in req'; } / 'in' __ 'opt' { return 'i
...
+} / 'query'  _ params:(lambda_param_decl_list)? _ ':=' _ query:table _ annotation:nl_annotation _ ';' {
+    if (annotation.name !== 'utterances') throw new Error('Only "utterances" annotation is allowed for examples');
+    return new Ast.Example('query', params ? params : [], query, annotation.value.value);
+} / 'action'  _ params:(lambda_param_decl_list)? _ ':=' _ action:action _ annotation:nl_annotation _ ';' {
+    if (annotation.name !== 'utterances') throw new Error('Only "utterances" annotation is allowed for examples');
+    return new Ast.Example('action', params ? params : [], action, annotation.value.value);
+} / 'program'  _ params:(lambda_param_decl_list)? _ ':=' _ program:program _ annotation:nl_annotation _ ';' {
+    if (annotation.name !== 'utterances') throw new Error('Only "utterances" annotation is allowed for examples');
+    return new Ast.Example('program', params ? params : [], program, annotation.value.value);
+}
```
gcampax: Do we not allow repeated annotations? I thought we agreed `#_[ann="one"] #_[ann="two"]` was equivalent to `#_[ann=["one", "two"]]`? Or not?

lib/prettyprint.js
```diff
@@ -287,11 +287,61 @@ function prettyprintProgram(ast, short) {
...
+        case 'action':
+            value = prettyprintAction(ex.value);
+            break;
+        case 'program':
+            value = prettyprintProgram(ex.value);
+            break;
+        default:
+            throw new TypeError();
+    }
+    let annotation = `#_[utterances=[${ex.utterances.map((u) => prettyprintValue(u)).join(',')}]]`;
```
gcampax: Is ex.utterances an Array of Ast.Value.String or an Array of JS string?
The grammar suggests the latter, in which case this code is incorrect: `prettyprintValue()` operates on `Ast.Value`

lib/prettyprint.js
```diff
@@ -287,11 +287,61 @@ function prettyprintProgram(ast, short) {
...
 function prettyprint(ast, short) {
     if (ast.isProgram)
         return prettyprintProgram(ast, short);
-    else
+    else if (ast.isPermissionRule)
         return prettyprintPermissionRule(ast, short);
+    else if (ast.isMeta)
+        return prettyprintMeta(ast, short);
+    else
+        return prettyprintDataset(ast, short);
```
gcampax: Wrong merge here? A dataset is not an input on its own, right?

lib/typecheck.js
```diff
@@ -771,8 +777,33 @@ async function typeCheckPermissionRule(permissionRule, schemas, getMeta = false)
...
 }
 
+async function typeCheckDataset(dataset, schemas, classes, getMeta = false) {
+    for (let ex of dataset.examples) {
+        await typeCheckDeclaration(ex, schemas, new Scope(), classes, getMeta, true);
+        if (!Array.isArray(ex.utterances))
+            throw new TypeError('Utterances annotation expects an array');
+        for (let utterance of ex.utterances) {
+            if (!utterance.isString)
+                throw new TypeError('Utterance can only be a string');
```
gcampax: Same here, is utterance a Ast.Value.String or a String?

test/test_typecheck.js
```diff
@@ -15,6 +15,7 @@ function typecheckTest() {
     var code = fs.readFileSync('./test/sample.apps').toString('utf8').split('====');
 
     Q.all(code.map((code) => {
+        if (code.indexOf(`// Meta language`) < 0) return Q();
```
gcampax: Left-over debugging code (and probably what's causing the coverage drop)

rayslxu: removed

lib/prettyprint.js
rayslxu: fixed

lib/ast.js
```diff
@@ -553,8 +553,33 @@ const Input = adt.data({
...
+    type: adt.only('stream', 'query', 'action', 'program'),
+    args: adt.only(Object),
+    value: (x, field, ctr) => {
+        adt.only(Stream, Table, Action, Input, Invocation)(x, field, ctr);
+        if (x instanceof Invocation)
+            return new Action.Invocation(x, x.schema);
+        else
+            return x;
+    },
+    utterances: adt.only(Array)
```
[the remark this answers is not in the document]
rayslxu: ID should be a string, right? for newly defined examples, what should the ID be?

lib/ast.js
```diff
@@ -553,8 +553,33 @@ const Input = adt.data({
...
+const Example = adt.newtype({
+    type: adt.only('stream', 'query', 'action', 'program'),
+    args: adt.only(Object),
+    value: (x, field, ctr) => {
+        adt.only(Stream, Table, Action, Input, Invocation)(x, field, ctr);
+        if (x instanceof Invocation)
+            return new Action.Invocation(x, x.schema);
+        else
+            return x;
+    },
```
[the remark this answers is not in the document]
rayslxu: Removed. But we actually never use `Ast.Action` in our grammar, so I added a line to convert the `action` to `Ast.Action.Invocation(action, action.schema)` in grammar for examples

lib/ast.js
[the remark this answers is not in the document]
rayslxu: Added ID and preprocessed, which default to -1 and [] for new examples.

lib/grammar.pegjs
rayslxu: I think this is just a design choice. I would like to keep the syntax I have unless there are some benefits to make the change.

lib/prettyprint.js
rayslxu: It should be JS string. I made a mistake. It's fixed now.

lib/typecheck.js
rayslxu: fixed.

lib/grammar.pegjs
rayslxu: What kind of metadata do we need? size, description maybe, contributors, or device used?

gcampax: Yeah it could be that too (see also https://github.com/Stanford-Mobisocial-IoT-Lab/almond-cloud/issues/24 ). 
I would keep it open ended though, like we do for functions.

lib/ast_api.js
```diff
@@ -340,7 +347,34 @@ PermissionRuleProto.iterateSlots = function* iterateSlots() {
...
+};
+
+ExampleProto.iterateSlots = function* iterateSlots() {
+    yield* this.value.iterateSlots();
+};
+
+MetaProto.iterateSlots = function* iterateSlots() {
+    for (let dataset of this.datasets)
+        yield* dataset.iterateSlots();
+};
```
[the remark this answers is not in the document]
rayslxu: Added.

lib/ast_api.js
```diff
@@ -340,7 +347,34 @@ PermissionRuleProto.iterateSlots = function* iterateSlots() {
...
+
+DatasetProto.iteratePrimitives = function* iteratePrimitives(includeVarRef) {
+    for (let ex of this.examples)
+        yield* ex.iteratePrimitives(includeVarRef);
+};
+
+ExampleProto.iteratePrimitives = function* iterateExample(includeVarRef) {
+    yield* this.value.iteratePrimitives(includeVarRef);
+};
```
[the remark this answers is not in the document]
rayslxu: I agree to generate synthesized programs. Added Example.toProgram.

lib/grammar.pegjs
rayslxu: so just an object and leave it empty for new dataset?

lib/ast.js
```diff
@@ -553,8 +553,33 @@ const Input = adt.data({
...
+    utterances: adt.only(Array)
+});
+module.exports.Example = Example.seal();
```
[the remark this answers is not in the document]
rayslxu: As we discussed, we will change declaration for commandpedia only. I will check how much they overlap after finishing the declaration and maybe create a base class for them if needed.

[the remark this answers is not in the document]
gcampax: Sounds good. Let's merge this now and deal with the base class later (as soon a swe find why the test coverage is still going down...)

lib/ast_api.js
```diff
@@ -76,7 +76,7 @@ ExampleProto.toProgram = function toProgram() {
         program = new Ast.Program([], [], [new Ast.Statement.Command(null, [this.value.clone()])], null);
     else if (this.type === 'query')
         program = new Ast.Program([], [], [new Ast.Statement.Command(this.value.clone(), [notifyAction()])], null);
-    else if (this.tyep === 'stream')
+    else if (this.type === 'stream')
         program = new Ast.Program([], [], [new Ast.Statement.Rule(this.value.clone(), [notifyAction()])], null);
```
gcampax: Good thing we added tests cause I had overlooked this

PR #26, opened 2018-10-04
Fix two tiny bugs

gcampax:


gcampax:
This is just bug fixes, I'm going to merge this without review

PR #51, opened 2018-11-06
Remove unused or obsolete code

gcampax:
This helps me figure out what is and is not needed for Genie


gcampax:
I'm 100% sure we will :slightly_frowning_face:

PR #67, opened 2019-02-01
lowering: support new interface of messaging

gcampax:
If messaging exposes the "getSelf" method, call it to obtain
the account to use to lower "return", passing the principal where
the program will be sent.

This allows the messaging interface to choose the correct return
address, based on who is executing the program.

This is related to https://github.com/Stanford-Mobisocial-IoT-Lab/thingengine-core/pull/42

### DIFF
--- a/lib/ast/api.js
+++ b/lib/ast/api.js
@@ -523,8 +523,11 @@ function makeToken(state) {
     return state.token++;
 }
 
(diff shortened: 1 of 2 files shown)


gcampax:
Yeah this patch needs to go in first, before the tests in thingengine-core can pass

PR #69, opened 2019-02-08
Wip/formatter

### LINKED ISSUE #10: formatter: handle projections correctly
This is a hack idea:
If a parameter is projected out, ignore all messages that use that parameter.

Eg, given spotify:
[ "The danceability of this song is ${danceability:%}.", "This song is in ${key} ${modifier}" ]
if you project on danceability you get the first message only.
If you project on key and modifier, you get the second only.

rayslxu:
A better formatter. fix #10

### DIFF
--- a/lib/runtime/formatter.js
+++ b/lib/runtime/formatter.js
@@ -95,7 +95,7 @@ module.exports = class Formatter extends FormatUtils {
     }
 
     _isNull(value) {
-        if (value === undefined || value === null || value === [])
+        if (value === undefined || value === null || (Array.isArray(value) && value.length === 0))
             return true;
         if (value instanceof Date && isNaN(value))
             return true;
@@ -225,7 +225,7 @@ module.exports = class Formatter extends FormatUtils {
             formatted = [formatted];
 
         // filter out null/undefined in the array
-        let filtered = formatted.filter((formatted) => !!formatted);
+        let filtered = formatted.filter((formatted) => !this._isNull(formatted));
         // flatten formatted (returning array in function causes nested array)
         return [].concat.apply([], filtered);
     }
--- a/lib/runtime/formatter.js
+++ b/lib/runtime/formatter.js
@@ -172,28 +172,36 @@ module.exports = class Formatter extends FormatUtils {
     }
 
     format(formatspec, argMap, hint) {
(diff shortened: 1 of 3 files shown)


lib/runtime/formatter.js
```diff
@@ -66,17 +68,41 @@ module.exports = class Formatter extends FormatUtils {
...
+            const param = match[1] || match[2];
+            if (!this._isNull(argMap[param]))
+                existSome = true;
+            match = reg.exec(str);
+        }
+        return !existSome;
+    }
+
+    _isNull(value) {
+        if (value === undefined || value === null || value === [])
```
gcampax: `value === []` is buggy, as you're comparing objects by address so it will always be false. Maybe `Array.isArray(value) && value.length === 0` ?

lib/runtime/formatter.js
```diff
@@ -179,23 +213,30 @@ module.exports = class Formatter extends FormatUtils {
...
+            return [];
         if (typeof formatted === 'object' &&
             formatted.type === 'text')
-            return formatted.text;
+            return [formatted.text];
         if (!Array.isArray(formatted))
             formatted = [formatted];
 
+        // filter out null/undefined in the array
+        let filtered = formatted.filter((formatted) => !!formatted);
```
gcampax: !! will also throw out the numbers 0 and NaN. Not sure if that's a problem, as both should be converted to strings, but something to keep in mind.

test/test_formatter.js
```diff
@@ -204,6 +204,65 @@ Picture: http://example.com/security-camera.jpg`
     }, null,
     [ 'The total file size is 7.' ]
     ],
+
+    ['com.wikicfp:search', {
+        start: new Date('TBD'),
+        end: new Date('TBD'),
```
gcampax: By the way, passing arbitrary stuff in the Date constructor is risky. I think the implementation of wikicfp should detect TBD and pass a null/NaN explicitly.

gcampax:
Also about the last point, I came across this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_date

Basically, in some not-so-well-defined cases, passing junk to `new Date` can throw instead of creating an invalid date.

PR #71, opened 2019-02-11
Add a tool to generate the grammar for genie-parser

### LINKED ISSUE #70: Add a tool to generate the grammar for the parser
Convert from the JS declaration in nn_grammar into a loadable Python module, or even a prebuilt SLR parser (so we don't duplicate the parser generator)

gcampax:
Take the existing parser generator tool, clean it up, polish it for external consumption, and give it a python backend.

Fixes #70


gcampax:
This PR fixes #70

PR #94, opened 2019-05-20
Wip/implicit casting

AydanPirani:
If I'm not mistaken, this should solve the issue with casting from numbers to currency.

### DIFF
--- a/lib/compiler/utils.js
+++ b/lib/compiler/utils.js
@@ -102,7 +102,7 @@ function compileCast(irBuilder, reg, type, toType) {
     
     if (type.isNumber && toType.isCurrency) {
         led casted = irBuilder.allocRegister();
-        compileBinaryOp(); //Fix up arguments to be passed into this function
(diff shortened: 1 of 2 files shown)


AydanPirani:
This should solve the implicit casting from numbers to currency.

gcampax:
Hi Aydan, thanks for your PR!

Please do not use the Github web editor, as that forces you to create one commit per file, which is not meaningful. You should have one commit per feature, or self-sufficient portion of the feature.
Please write meaningful commit messages as well. Your commit message should have a subject in the first line, containing the area you're modifying and what you're doing. You should have body (separated from the subject with one empty line) indicating what you're doing, why you're doing (what problem you're trying to solve) and how you're doing it. See for example https://code.likeagirl.io/useful-tips-for-writing-better-git-commit-messages-808770609503 for useful guidelines.

I have not looked at the code yet, as it seems tests are not passing quite yet. Please fix the tests and after that I will review the PR. Alternatively, let me now if you struggle to understand a test failure, and I'll take a look.

PR #95, opened 2019-05-26
WIP/Implicit-Casting

AydanPirani:
Checked if the conversion between number and currency is valid, then called a unary operation to cast between the two types


gcampax:
Why did you open a second pull request? You already had #94 open.

AydanPirani:
> Why did you open a second pull request? You already had #94 open.

At the time, I did not realize that there was another pull request existing. I have closed the other pull request.

gcampax:
Anyway, what is missing now is just the implementation of `get_currency`, which should live in `lib/builtin/primitive_ops.js`.

Also, as Coveralls notes, you should add some tests. Typecheck tests are defined in tests/sample.apps and run with `test/test_typecheck.js`. Compiler tests are in `test/test_compiler.js` and runtime tests are in `test/test_runtime.js`. You will probably need all three of them to make sure this is implemented correctly.

AydanPirani:
> I just realized that you also need to add the definition of this primitive op in lib/builtin/defs.js
> 
> Anyway please add a test, so as to exercise this code and actually verify it is correct.
> (It is not, in fact, as the linter noted)

I have added in the definition of get_currency and committed these changes. Would I add the test cases in test_builtin?

PR #111, opened 2019-07-19
Use .tt instead of .json

rayslxu:


rayslxu:
I have merged the latest changes. Add your spotify changes to thingpedia.tt should make the test pass @gcampax

PR #113, opened 2019-07-22
Wip/sparql converter

ryachen01:
Helper function to convert from ast table to sparql query for wikidata


lib/helper.js
```diff
@@ -0,0 +1,12 @@
...
+const SparqlConverter = require("./sparql_converter");
+
+module.exports = {
+    program_to_sparql(program){
+        return SparqlConverter.program_to_sparql(program);
+    },
+    table_to_sparql(table){
+        return SparqlConverter.table_to_sparql(table);
+    },
+};
```
rayslxu: I think it would be better to merge these two with one "toSparql" exposed to users. You can determine the type by `isTable` and `isProgram`. Note that, programs can potentially have no table in it. Although in practice we won't do that, but we should avoid using things like `program["rules"][0]["table"]["table"]` without any protection

gcampax:
Hi Ryan, please do not add any dependency to thingtalk. The library must be loadable in the browser (through browserify/webpack), hence node-specific modules like `node-fetch` are not allowed.

test/test_sparql_converter.js
```diff
@@ -0,0 +1,107 @@
...
+
+                        const queryURL =
+                            "https://query.wikidata.org/sparql" +
+                            "?query=" +
+                            encodeURIComponent(sparqlQuery[0]);
+                        resolve(queryURL);
+                    }
+                );
+            });
+            return promise;
```
gcampax: This promise code seems to be needlessly convoluted. Why do you need to create a `new Promise`?

test/test_sparql_converter.js
```diff
@@ -0,0 +1,107 @@
...
+    ];
+
+    Promise.all(
+        code.map((code) => {
+            let promise = new Promise((resolve, reject) => {
+                code = code.trim();
+                AppGrammar.parseAndTypecheck(code, _schemaRetriever).then(
+                    async (program) => {
+                        //convert from ast to sparql
+                        const sparqlQuery = await SparqlConverter.toSparql(
```
gcampax: Is `toSparql` async? Why?

lib/sparql_converter.js
```diff
@@ -0,0 +1,386 @@
...
+                        sort_triplet,
+                        order_results
+                    );
+                }
+            }
+        }
+    },
+
+    //converts filter to triplet
+    filter_to_triplet(filter_input, schema_input, index) {
```
gcampax: How do you handle comparison operators (`>=`, '<=`)?

lib/sparql_converter.js
```diff
@@ -0,0 +1,386 @@
...
+                false,
+            ];
+        } else if (`${type}` === "Number" || `${type}`.includes("Measure")) {
+            let formatted_value = `"%s"^^xsd:decimal`.format(value);
+            return [no_label_triplet.format(parameter, formatted_value), false];
+        } else if (`${type}`.includes("Entity")) {
+            return [no_label_triplet.format(parameter, "wd:" + value), false];
+        } else {
+            return [
+                get_label_triplet.format(index, value),
```
gcampax: How do you handle String-type parameters vs Entity parameters? Do you need escaping?

lib/sparql_converter.js
```diff
@@ -0,0 +1,386 @@
...
+const get_projection_triplet = `?v1 wdt:%s ?v3.\n`;
+
+//triplet required if there is a projection
+const sort_results_triplet = `?v%s wdt:%s ?counter.\n`;
+
+//for ordering results either ascending or descending
+const order_template = `ORDER BY %s(?counter)`;
+
+module.exports = {
+    toSparql(input) {
```
gcampax: This is very much Wikidata specific, and the name should reflect it.
Also, given that there is only one wikidata, I wonder what's the advantage of having this in the main thingtalk library...

PR #114, opened 2019-07-24
Add a method to return the canonical of arguments

rayslxu:
Handle both string and object canonical annotations

### DIFF
--- a/lib/ast/function_def.js
+++ b/lib/ast/function_def.js
@@ -59,6 +59,15 @@ class ArgumentDef {
         this.unique = this.annotations.unique && this.annotations.unique.isBoolean && this.annotations.unique.value === true;
     }
 
+    get canonical() {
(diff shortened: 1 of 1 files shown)


lib/ast/function_def.js
```diff
@@ -190,6 +190,17 @@ class ExpressionSignature {
...
+    getArgCanonical(arg) {
+        const a =this._argmap[arg];
+        if (a) {
+            if (typeof a.canonical === 'string')
+                return a.canonical;
+            if (typeof a === 'object' && 'npp' in a)
+                return a['npp'];
+        }
+        return arg;
+    }
```
gcampax: Would this be better as `.canonical` or `.getCanonical` on the argument object?
Also, I thought we did not want to hardcode `npp` because that's English-specific?

PR #121, opened 2019-07-30
Allow predefined times when converting to nn-syntax

rickygv99:


gcampax:
You will need to change also the parser in lib/nn-syntax/parser.lr, and because you're adding a new special kind of token probably lib/nn-syntax/lexer.js Also I'm not sure the syntax you have chosen is that great, given that you collapsed everything into one token...

lib/nn-syntax/parser.lr
```diff
@@ -431,7 +431,12 @@ constant_Date = {
     date:constant_Date '-' offset:constant_Measure_ms => new Ast.Value.Date(date.value, '-', offset);
 }
 
-constant_Time = time:TIME => new Ast.Value.Time(time.value.hour, time.value.minute, time.value.second||0);
+// Note that while keeping digits in ranges [0-23],[0-59],[0-59] is not enforced
+// here, it is enforced in tonn_converter, rendering it unnecessary to do so here
+constant_Time = {
+    time:TIME => new Ast.Value.Time(time.value.hour, time.value.minute, time.value.second||0);
+    'TIME:' hour:LITERAL_INTEGER ':' minute:LITERAL_INTEGER ':' second:LITERAL_INTEGER => new Ast.Value.Time(hour, minute, second);
```
gcampax: This syntax expects 6 tokens: "TIME:", a literal integer, ':', another literal integer, ':', and a third literal integer
(this is a parser definition, so every element is either a non-terminal or a token)

OTOH, your toNN code generates a single token. Tokens are separated by spaces in NN syntax, and if you want multiple tokens you need to use the `List` syntax.

Regardless, hardcoding the numbers in the code is not a great idea, because:
- now you will need explicit training data for all sorts of combinations
- the user gets no control over what "morning" or "afternoon" means, and it's all hardcoded deep in the library
 in reality, there is are a finite small number of parts of day that are interesting, so it would be better to model those and let the dialog agent sort it out, similar to what we do for relative locations

rickygv99: I now represent the time with a single token: LITERAL_TIME, which I added to the lexer.

I'll look into the implementation of relative locations and see if I can pull some of that logic into how I handle times. I might keep a couple numbers hardcoded still though, such as noon and midnight, so the LITERAL_TIME token should still prove useful in at least those cases.

I was planning on hardcoding the numbers in timers.genie and mapping different phrases onto them rather than implementing phrases such as "in the morning" directly into the thingtalk grammar -- if I do that, would that still require explicit training data for combinations of those phrases and other sentence fragments?

gcampax:
Ok, this is correct now (you can verify it by adding more tests to test_nn_syntax.js, which you should do anyway), but still, it does not strike me as a great idea to just hardcode what morning means...

lib/nn-syntax/parser.lr
gcampax: You'll need explicit training data to understand "in the morning" no matter what. The difference is whether you need training data for `12:00:00`, `12:00:01`, `12:00:02` etc...

rickygv99:
I added Absolute and Relative Times, with possible values of "morning" and "evening" for the relativeTag in Time.Relative.

PR #123, opened 2019-08-05
describe: fix timers with undefined parameters

gcampax:
We cannot assume that the time parameter will be an actual Time
value and not some other kind of value.

### DIFF
--- a/lib/describe.js
+++ b/lib/describe.js
@@ -630,13 +630,11 @@ class Describer {
             const times = stream.time.map((t) => this.describeArg(t));
 
             if (stream.expiration_date === null) {
-              return this._("every day %s %s").format(
-                  stream.time[0].value.isAbsolute ? "at" : "in",
+              return this._("every day at %s").format(
                   times.length > 1 ? times.reduce((t1, t2) => this._("%s and %s").format(t1, t2)) : times[0]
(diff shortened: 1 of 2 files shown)


rickygv99:
I think we still need to check whether the value is absolute or not -- this determines whether we should use "in" or "at" in our response: (e.g. "[in] the morning" vs "[at] 4:00 PM"). We can fix the issue of timers with undefined parameters while retaining this check by instead replacing the line `stream.time[0].value.isAbsolute ? "at" : "in"` with the line `(stream.time[0] instanceof Ast.Value.Time) && (stream.time[0].value.isAbsolute === false) ? "in" : "at"`.

I tested this change and it passes both the old tests and the newly added test. If you think it looks good, then I'll commit it.

gcampax:
Honestly, I'd rather live with slightly less grammatical English (which can/should be fixed somewhere else with a string replace) than adding more hacks here. Not to mention that the hacks make the strings impossible to translate meaningfully...

rickygv99:
Sounds good -- we can replace the grammar elsewhere then. If we remove the grammatical change in describe.js, we should adjust the corresponding test to say "at the morning" rather than "in the morning" and then it's good to go.

PR #126, opened 2019-08-09
changed query output variable to projection parameter

ryachen01:


test/test_sparql_converter.js
```diff
@@ -24,14 +24,15 @@ const TEST_CASES = [
...
     ?v2 wdt:P734 ?item1.
     ?v2 wdt:P641 wd:Q5372.
-    ?v2 wdt:P22 ?v1.
+    ?v2 wdt:P22 ?P22.
     SERVICE wikibase:label { bd:serviceParam wikibase:language "en". }
     }
-    limit 10`,
+    limit 10
+    `,
     ],
```
rayslxu: You can do the following such that it actually returns `P22` instead of `P22Label`:
```
SELECT DISTINCT ?P22 WHERE{
    ?item1 ?label 'Curry'@en.
    ?v2 wdt:P734 ?item1.
    ?v2 wdt:P641 wd:Q5372.
    ?v2 wdt:P22 ?v3.
    SERVICE wikibase:label { bd:serviceParam wikibase:language "en". ?v3 rdfs:label ?P22}
    }
LIMIT 10
```

test/test_sparql_converter.js
```diff
@@ -137,14 +138,14 @@ const TEST_CASES = [
     join ([id] of @org.wikidatasportsskill.sports_team())), P647 == lhs.id && P286 == rhs.id => notify;
         `,
         `
-    SELECT distinct ?v1Label WHERE{
```
rayslxu: Why are other examples above not affected? Shouldn't they all be using id instead of v1?

PR #128, opened 2019-08-10
added support for substrings

ryachen01:
added support for substrings and fixed a minor bug


lib/wikidata_sparql.js
```diff
@@ -587,6 +593,7 @@ module.exports = {
         let additionalCode = "";
 
         let requiredShift = 0;
+        let adjustedShift = false;
```
gcampax: Sorry this is getting a little hard to follow, would you mind adding comments explaining the meaning of these variables and the logic here? Thanks!

PR #132, opened 2019-08-14
Wip/sparql cleanup

ryachen01:


lib/wikidata_sparql2.js
```diff
@@ -10,81 +10,119 @@
...
-        assert(program.isProgram && program.rules.length === 1 && program.rules[0].table);
+        this._nextTableIndex = 0;
+        this._tableIndices = new Map(); // from Invocation to number
+
+        this._nextParamId = 0;
+        this._boundParameters = new Set();
+
+        this._whereClause = [];
+        this._additionalSparql = [];
+        this._stringConversions = [];
```
rayslxu: You can probably remove the same code in constructor

lib/wikidata_sparql2.js
```diff
@@ -117,61 +166,170 @@ module.exports = class WikidataSparqlConverter {
...
+            this._additionalSparql.push(sort_code);
+            return [scope, projection];
         } else if (Utils.isUnaryTableToTableOp(table)) {
             return this._tableToSubquery(table.table);
         } else {
             throw new TypeError(`Not implemented`);
         }
     }
 
+    reformat_date(date) {
```
rayslxu: Starts the function name with "_" if it won't be used outside the class. Or even better, for this one, it can just be a standalone function since it does not touch the class at all.

PR #142, opened 2019-09-12
Extend thingtalk to handle schema.org

rayslxu:
- Add compound value type

### DIFF
--- a/lib/ast/function_def.js
+++ b/lib/ast/function_def.js
@@ -99,8 +99,8 @@ class ArgumentDef {
             metadata, annotations);
     }
 
-    toString() {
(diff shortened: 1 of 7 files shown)


lib/ast/function_def.js
```diff
@@ -314,17 +351,19 @@ class FunctionDef extends ExpressionSignature {
     toString(prefix = '') {
         let annotations = prettyprintAnnotations(this);
         const firstline = `${prefix}${this.is_monitorable ? 'monitorable ' : ''}${this.is_list ? 'list ' : ''}${this.functionType} ${this.name}`;
+        // skip arguments flattened from compound param
+        const args = this.args.filter((a) => !a.includes('.'));
```
rayslxu: this hack to skip flattened arguments is a little bit ugly

PR #144, opened 2019-09-19
Add syntax for subfunctions

rayslxu:


lib/ast/function_def.js
```diff
@@ -350,6 +350,47 @@ class FunctionDef extends ExpressionSignature {
...
+    _checkConflictArguments(functions) {
+        const args = {};
+        for (let f of functions) {
+            for (let a of f.args) {
+                if (!(a in args)) {
+                    args[a] = f._argmap[a];
+                    continue;
+                }
+                const err_msg = `Parameter ${a} is defined multiple times in ${functions.map((f) => f.name).join(', ')}`;
+                if (args[a].canonical !== f.getArgCanonical(a))
```
gcampax: Did you mean to call getArgCanonical on both argument? Or do deep equality here?

lib/ast/function_def.js
```diff
@@ -350,6 +350,47 @@ class FunctionDef extends ExpressionSignature {
...
+        args = args.concat(this.args.map((a) => this._argmap[a]));
+        functions.forEach((f) => {
+            f.args.forEach((a) => {
+                if (!this.args.includes(a))
+                    args.push(f._argmap[a]);
+            });
+        });
+
+        return new FunctionDef(this._functionType, this._name, args,
+            this.is_list, this.is_monitorable, metadata, annotations, this._parent);
```
gcampax: Do we need to check is_list/is_monitorable for compatibility with the base functions?

lib/grammar.pegjs
```diff
@@ -228,10 +249,16 @@ import_mixin_stmt = 'import' __ first:ident _ rest:(',' _ ident _)* 'from' __ na
   return new Ast.ImportStmt.Mixin([first].concat(take(rest, 2)), name, in_params);
 }
 
-function_def = modifiers:query_modifiers? _ type:('query' / 'action') __ name:ident _ args:function_param_decl_list _ annotations: ( _ annotation _)* ';' _ {
+function_name_list = first:ident _ rest:(',' _ ident _)* {
+    return [first].concat(take(rest, 2))
+}
+function_def = modifiers:query_modifiers? _ type:('query' / 'action') __ name:ident _ extends_:('extends' _ function_name_list _ )? args:function_param_decl_list _ annotations: ( _ annotation _)* ';' _ {
```
gcampax: You must use `__` after `'extends'` otherwise `extendsfoo` would parse as `extends foo`

lib/ast/function_def.js
rayslxu: `args[a]` is `ArgumentDef`, and i didn't keep track of function information for simplicity

lib/ast/function_def.js
rayslxu: no, I thought about this. I think it's reasonable for is_list/is_monitorable to be different from base functions. e.g., I can extend a monitorable base function with a parameter which is random, it will become not monitorable.

gcampax: Yeah but the other way around doesn't make sense? And also `list` should be the same (either list or non-list)?

PR #148, opened 2019-09-23
Changes to support schema.org queries and actions

rayslxu:


rayslxu:
@gcampax This version should work with the current code base, maybe we can merge this?

PR #162, opened 2019-11-01
Extend device selectors to refer to devices by name

gcampax:
So we can ask "turn on my bedroom lights"

This is a feature asked by @balloob.
This is only part 1. Part 2 is the dialog agent to actually select the device. Part 3 is Genie to train those sentences. Part 4 is every Thingpedia device that would like this functionality should add relevant templates and change their confirmation string.

### DIFF
--- a/lib/ast/program.js
+++ b/lib/ast/program.js
@@ -27,48 +27,76 @@ const {
  * Selectors correspond to the `@`-device part of the ThingTalk code,
  * up to but not including the function name.
  *
- * @class
  * @alias Ast.Selector
+ * @property {boolean} isSelector - true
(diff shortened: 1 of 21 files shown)


lib/ast/program.js
```diff
@@ -27,48 +27,84 @@ const {
...
+
+        this.all = all;
+    }
+
+    clone() {
+        return new DeviceSelector(this.kind, this.id, this.principal, this.attributes.slice());
+    }
+
+    toString() {
+        return `Device(${this.kind}, ${this.id ? this.id : ''}, )`;
```
rayslxu: no attributes?

PR #173, opened 2019-12-01
Fix typecheck of input args

rayslxu:
If the value of an input is undefined, do not remove it from the schema

### DIFF
--- a/lib/typecheck.js
+++ b/lib/typecheck.js
@@ -667,6 +667,9 @@ function typeCheckInputArgs(ast, schema, scope, classes) {
         if (!inParamType || !schema.isArgInput(inParam.name))
             throw new TypeError('Invalid input parameter ' + inParam.name);
 
+        if (inParam.value.isUndefined)
+            continue;
+
         const valueType = typeForValue(inParam.value, scope);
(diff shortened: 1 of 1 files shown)


rayslxu:
> Undefined means the parameter is set (to be slot-filled) so it should be removed from the schema, otherwise you can parameter pass into it, which is wrong. What is this solving? Do you have a test case?

In the make rule GUI, I need to get the input arguments.

gcampax:
Why do you need to go through typecheck in make rule? Presumably you have an `Invocation` directly, not an Ast.Table or Ast.Action? Or maybe you should use the primitive templates instead, like we do in Android? In any case, I'm not convinced this patch is correct.

rayslxu:
> Why do you need to go through typecheck in make rule? Presumably you have an `Invocation` directly, not an Ast.Table or Ast.Action? Or maybe you should use the primitive templates instead, like we do in Android? In any case, I'm not convinced this patch is correct.

I am using the primitive templates, that's why I don't have invocation directly. I guess I can do iteratePrimitives, and track what parameter has been used locally...

gcampax:
Wait no, if you're using the primitive templates, then you have the primitive template arguments, and you're good to go. Beta reduce those, do nothing else.

PR #179, opened 2020-01-05
Mark `filter` as a keyword

gcampax:
We don't want people accidentally using it as identifier, because
that would be ambiguous with the `filter` keyword already in use.

Instagram has been updated in Thingpedia already.


gcampax:
You mean the paraphrases? Yeah I did.

PR #184, opened 2020-02-13
Fix compiler undefined devices bug

euirim:
Fix problem where compiler returns an obscure error when device and handle_thingtalk is undefined.

### DIFF
--- a/lib/compiler/ast-to-ops.js
+++ b/lib/compiler/ast-to-ops.js
@@ -164,7 +164,10 @@ function compileTableToOps(table, extra_in_params) {
         throw new NotImplementedError(table);
 
     if (table.isVarRef) {
-        return new TableOp.InvokeVarRef(table.name, table.in_params.concat(extra_in_params), table);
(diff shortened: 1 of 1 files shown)


lib/compiler/ast-to-ops.js
```diff
@@ -164,7 +164,10 @@ function compileTableToOps(table, extra_in_params) {
         throw new NotImplementedError(table);
 
     if (table.isVarRef) {
-        return new TableOp.InvokeVarRef(table.name, table.in_params.concat(extra_in_params), table);
+	const compiled = new TableOp.InvokeVarRef(table.name, table.in_params.concat(extra_in_params), table);
+        compiled.device = null;
+        compiled.handle_thingtalk = false;
+        return compiled;
     } else if (table.isResultRef) {
         return new TableOp.ReadResult(table.kind + ':' + table.channel, table.index, table.schema, table);
```
gcampax: This line here has the same problem, can you fix that at the same time?
Also, there is an indent problem.

euirim: Fixed indent and took a stab at fixing the second instance.

PR #188, opened 2020-02-27
My changes from the Multiwoz branch

gcampax:
All changes that are not dialogue related


lib/builtin/defs.js
```diff
@@ -124,6 +114,28 @@ module.exports.BinaryOps = {
         fn: 'contains',
         flip: true
     },
+    'contains~': {
```
rayslxu: can we have comments with examples here?

lib/ast/expression.js
```diff
@@ -386,6 +386,18 @@ class BooleanExpression extends Node {
 }
 BooleanExpression.prototype.isBooleanExpression = true;
 
+function arrayEquals(a1, a2) {
+    if (a1 === a2)
+        return true;
+    if (a1.length !== a2.length)
+        return false;
+    for (let i = 0; i < a1.length; i++) {
```
rayslxu: so two arrays have to be in the same order? we don't normalize during generation in genie right?

gcampax: We don't, no, so this is a very strict form of equality. In practice in Genie I only need equality of Atom and Not(Atom), so we can relax it later if we need it.

PR #193, opened 2020-03-19
Add .equals() to Ast.Value classes

gcampax:
As was previously provided by adt


gcampax:
Yeah it's meant to be the same as the equals() provided by adt. And order matters in general - two programs that call the same function with arrays that differ in order in general will do different things. Only for `in_array` the order is irrelevant...

PR #198, opened 2020-04-18
Fixes in preparation to homework 2

gcampax:


gcampax:
Whoa LGTM caught an actual bug! It's a miracle!

PR #199, opened 2020-04-23
Bug fixes

rayslxu:
- Fix canonical interface for argument
- Fix tonn converter for locations

### DIFF
--- a/lib/ast/function_def.js
+++ b/lib/ast/function_def.js
@@ -173,8 +173,14 @@ class ArgumentDef extends Node {
         let canonical = this.nl_annotations.canonical;
         if (typeof canonical === 'string')
             return canonical;
-        if (typeof canonical === 'object' && 'npp' in canonical)
-            return canonical['npp'][0];
+        if (typeof canonical === 'object') {
+            if ('base' in canonical)
+                return canonical['base'][0];
+            if ('property' in canonical)
+                return canonical['property'][0];
+            if ('npp' in canonical)
+                return canonical['npp'][0];
+        }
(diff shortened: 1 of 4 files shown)


rayslxu:
> Can you elaborate on what exactly is the bug here? Why does the canonical matter? Especially given that we expand all canonical forms in Genie?

So when we encounter a filter of `count(x)`, we use the `canonical` to check if there is a parameter with the name `xCount`. 
Of course i can switch to use base form of canonical in my script, but given that we are not deprecate this method from thingtalk, i think we should update it here.


also, i'm adding other fixes here, so i will change the name of the pr

gcampax:
Yeah I think the ThingTalk changes are correct, but I also wish Genie did not depend on the specific of how we generate compatibility natural language annotations. Annotations are, by definition, something that ThingTalk shouldn't know about, and only meaningful to higher-level software.

Anyway the other patch looks good too.

rayslxu:
> Yeah I think the ThingTalk changes are correct, but I also wish Genie did not depend on the specific of how we generate compatibility natural language annotations. Annotations are, by definition, something that ThingTalk shouldn't know about, and only meaningful to higher-level software.
> 
> Anyway the other patch looks good too.

unfortunately, i have to rely on annotations because the paramter name is `aggregateRating.reviewCount`....


also, you get a free update for your dialog test, no more "tell me the id", but "tell me the name"

gcampax:
Yes, I get a free bug fix, minus the fact that that's the legacy slot filling code, combined with an $undefined that should never be generated :sweat_smile:

PR #217, opened 2020-07-06
Relax entity checks in NN syntax

gcampax:
As necessary to annotate MultiWOZ

Opening this PR so I don't forget about this branch, but there is some work needed before this can land.

### DIFF
--- a/lib/ast/values.js
+++ b/lib/ast/values.js
@@ -114,11 +114,11 @@ module.exports.Time = Time;
 class AbsoluteTime extends Time {
     constructor(hour, minute, second) {
         super();
-        assert(typeof hour === 'number');
+        assert(typeof hour === 'number' && Number.isFinite(hour));
         this.hour = hour;
-        assert(typeof minute === 'number');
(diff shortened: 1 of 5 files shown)


lib/ast/values.js
```diff
@@ -114,11 +114,11 @@ module.exports.Time = Time;
...
-        assert(typeof hour === 'number');
+        assert(typeof hour === 'number' && Number.isFinite(hour));
         this.hour = hour;
-        assert(typeof minute === 'number');
+        assert(typeof minute === 'number' && Number.isFinite(minute));
         this.minute = minute;
-        assert(typeof second === 'number');
+        assert(typeof second === 'number' && Number.isFinite(second));
         this.second = second;
     }
```
rayslxu, line 123: Shouldn't this be hour < 24, minute, second < 60?

gcampax: Yeah probably. Hour >= 0 too. I don't remember why I added this check though. I think it was to catch NaNs that show up as a result of math with `undefined`.

PR #220, opened 2020-07-13
Entity definition statements

gcampax:
This PR introduces a new "entity" statement inside a ThingTalk class.

The goal is to make classes self-contained, and eliminate the need for entities.json (except for well-known entities like tt:stock_id or tt:country).
This in particular solves a chicken-and-egg problem we have today: you cannot upload a device to Thingpedia that uses an entity type not defined in Thingpedia, but you cannot upload an entity to Thingpedia unless you own that device first. So you have to upload an empty placeholder device, upload the entity, and then upload the final device. It's convoluted and messy.
(String values have the same problem - for strings, we should demote the error to a warning)
It also helps with using Genie, as we can make the starter code download the standard entities from Thingpedia and not have to explain how to modify it.

Entity definitions have three properties:
- the name, which is concatenated with the class name to form the full entity ID
- NL annotations, which are used in Thingpedia and in some error messages from the agent
- impl annotations, the most important of which is `has_ner` which corresponds to the `has_ner_support` bit in entities.json, and indicates whether an entity can appear in the sentence or not; the name is not great but I could not think of a better name. Defaults to `true` if unspecified, which is almost always what you want, and not too harmful if incorrect.

### DIFF
--- a/lib/ast/class_def.js
+++ b/lib/ast/class_def.js
@@ -36,8 +36,9 @@ class ClassDef extends Statement {
      *        in the source code
      * @param {string} kind - the class identifier in Thingpedia
      * @param {Ast.ClassDef[]|null} _extends - parent classes (if any)
-     * @param {Object.<string, any>} members - the class members including queries, actions, imports
+     * @param {Object.<string, any>} members - the class members including queries, actions, entities, imports
      * @param {Ast.ImportStmt[]} [members.imports=[]] - import statements in this class
+     * @param {Ast.EntityDef[]} [members.entities=[]] - entity declarations in this class
      * @param {Object.<string, Ast.FunctionDef>} [members.queries={}] - query functions in this class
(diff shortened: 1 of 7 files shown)


gcampax:
> where do you check if an entity has been defined?

In ThingTalk I don't.
In Thingpedia, it's at https://github.com/stanford-oval/almond-cloud/blob/master/util/validation.js#L123

rayslxu:
> > where do you check if an entity has been defined?
> 
> In ThingTalk I don't.
> In Thingpedia, it's at https://github.com/stanford-oval/almond-cloud/blob/master/util/validation.js#L123

So there will be another pr for almond-cloud to update `validation` ?

gcampax:
>
> So there will be another pr for almond-cloud to update `validation` ?

Oh there will be a _lot_ of PRs for almond-cloud. Think of how much we changed Genie...

PR #232, opened 2020-07-31
Add DatePiece class

satojk:
DatePiece is a way to construct Date values when only one of [year,
month, day] is specified. The big-end non-supplied values get set
to the present and the little-end non-supplied values get set to the
smallest possible. For example: DatePiece('month', 3) gets compiled to
the Date corresponding to the first of April in the present year
(January is month 0). This is useful for disambiguating a reference
such as "the 11th" in the context of dates.

TODO: parser functionalities for DatePiece

### DIFF
--- a/lib/date_utils.js
+++ b/lib/date_utils.js
@@ -88,12 +88,32 @@ function createEdgeDate(edge, unit) {
     return date;
 }
 
+function createDatePiece(piece, value) {
+    const date = new Date;
+    date.setHours(0, 0, 0, 0);
+    switch(piece) {
+        case 'year':
+            date.setMonth(0, 1); // 1st of Jan
+            date.setYear(value);
+            break;
+        case 'month':
+            date.setMonth(value - 1, 1); // 1st of Jan, this year
+            break;
+        case 'day':
+            date.setDate(value); // this day in the month
+            break;
+    }
+    return date;
+}
+
 module.exports = {
     normalizeDate(value) {
         if (value === null)
             return new Date;
         else if (value instanceof Date)
             return value;
+        else if (typeof value.edge === 'undefined')
+            return createDatePiece(value.piece, value.value);
         else
             return createEdgeDate(value.edge, value.unit);
     },
--- a/lib/date_utils.js
+++ b/lib/date_utils.js
@@ -107,7 +107,7 @@ function createDatePiece(year, month, day, time) {
         date.setHours(0, 0, 0, 0);
     }
     if (time !== null)
-        date.setHours(time.hour, time.minute, 0, 0);
+        date.setHours(time.hour, time.minute, time.second);
     return date;
 }
 
--- a/lib/date_utils.js
+++ b/lib/date_utils.js
@@ -106,7 +106,7 @@ function createDatePiece(year, month, day, time) {
         date.setDate(day);
         date.setHours(0, 0, 0, 0);
     }
-    if (time !== -1)
+    if (time !== null)
         date.setHours(time.hour, time.minute, 0, 0);
     return date;
 }
--- a/lib/date_utils.js
+++ b/lib/date_utils.js
@@ -88,7 +88,7 @@ function createEdgeDate(edge, unit) {
     return date;
 }
 
-function createDatePiece(year, month, day, hours, minutes) {
+function createDatePiece(year, month, day, time) {
     // All non-supplied values to the left of the largest supplied
     // value are set to the present. All non-supplied values to the
     // right of the largest supplied value are set to the minimum.
@@ -106,10 +106,8 @@ function createDatePiece(year, month, day, hours, minutes) {
         date.setDate(day);
         date.setHours(0, 0, 0, 0);
     }
-    if (hours >= 0)
-        date.setHours(hours, 0, 0, 0);
-    if (minutes >= 0)
-        date.setHours(date.getHours(), minutes, 0, 0);
+    if (time !== -1)
+        date.setHours(time.hour, time.minute, 0, 0);
     return date;
 }
 
@@ -120,7 +118,7 @@ module.exports = {
         else if (value instanceof Date)
             return value;
         else if (typeof value.edge === 'undefined')
-            return createDatePiece(value.year, value.month, value.day, value.hours, value.minutes);
+            return createDatePiece(value.year, value.month, value.day, value.time);
         else
             return createEdgeDate(value.edge, value.unit);
     },
--- a/lib/date_utils.js
+++ b/lib/date_utils.js
@@ -88,21 +88,28 @@ function createEdgeDate(edge, unit) {
     return date;
 }
 
-function createDatePiece(piece, value) {
+function createDatePiece(year, month, day, hours, minutes) {
+    // All non-supplied values to the left of the largest supplied
+    // value are set to the present. All non-supplied values to the
+    // right of the largest supplied value are set to the minimum.
     const date = new Date;
-    date.setHours(0, 0, 0, 0);
-    switch(piece) {
-        case 'year':
-            date.setMonth(0, 1); // 1st of Jan
-            date.setYear(value);
-            break;
-        case 'month':
-            date.setMonth(value - 1, 1); // 1st of Jan, this year
-            break;
-        case 'day':
-            date.setDate(value); // this day in the month
-            break;
+    if (year > 0) {
+        date.setYear(year);
+        date.setMonth(0, 1); // 1st of Jan
+        date.setHours(0, 0, 0, 0);
     }
+    if (month > 0) {
+        date.setMonth(month - 1, 1);
+        date.setHours(0, 0, 0, 0);
+    }
+    if (day > 0) {
+        date.setDate(day);
+        date.setHours(0, 0, 0, 0);
+    }
+    if (hours >= 0)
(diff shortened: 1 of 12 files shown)


lib/nn-syntax/parser.lr
```diff
@@ -747,6 +748,13 @@ constant_Date = {
 
     'start_of' unit:UNIT => new Ast.Value.Date(new Ast.DateEdge('start_of', unit.value));
     'end_of' unit:UNIT => new Ast.Value.Date(new Ast.DateEdge('end_of', unit.value));
+
+    'new' 'DatePiece' '(' year:literal_number ',' month:literal_number ','
+                          day:literal_number ',' hours:literal_number ','
+                          minutes:literal_number ')' => {
```
gcampax, line 754: this needs to be a constant_Number, not a literal_number, to pick up NUMBER_ and similar tokens

test/test_nn_syntax.js
```diff
@@ -618,6 +618,17 @@ now => @org.schema.restaurant() => notify;`],
      `get sunrise sunset on the NUMBER_0 th`, { NUMBER_0: 25 },
      `now => @org.thingpedia.weather.sunrise(date=new DatePiece(-1, -1, 25, -1)) => notify;`],
 
+    [`now => @org.thingpedia.weather.sunrise param:date:Date = new DatePiece ( - 1 , - 1 , 10 , TIME_0 ) => notify`,
```
rayslxu, line 621: do we have to use space between `-` and `1` here?  It looks weird to me

satojk: From what I could tell, negative numbers in NN Syntax are always separated from the `-`, such as [in here](https://github.com/stanford-oval/thingtalk/blob/master/lib/nn-syntax/tonn_converter.js#L264), so I kept it like that. Don't know if there's a special reason for it.
Cc @gcampax

rayslxu: l talked to giovanni. let's not put -1 in the syntax.

rayslxu: let's do `now => @org.thingpedia.weather.sunrise param:date:Date = new DatePiece ( , , 10 , TIME_0 ) => notify`

rayslxu: or `now => @org.thingpedia.weather.sunrise param:date:Date = new DatePiece ( null , null , 10 , TIME_0 ) => notify` if you prefer

satojk: Ok, as per my last commit now, the NN Syntax omits the non-supplied values (`new DatePiece ( , , 10 , TIME_0 )`).

In the ThingTalk and JS syntaxes, it's taking in `null` instead of `-1` for non-supplied (`new DatePiece(null, null, 10, time)`), But thinking now, maybe I misinterpreted and you had meant for these to stay the same with `-1`. Is it a bad thing that accidental `null`s would get interpreted the same as the intentional non-supplied args? Would be very quick to rollback this change, so let me know and I'll do it immediately.

gcampax: Oh, my take is there is no such thing as "accidental `null`". If "null" is not a valid value in a field, it's a programmer error (and we can put an assertion to catch it). If "null" is a valid value, we give it a meaning.

gcampax: Also, please make ThingTalk surface syntax consistent with NN syntax, unless there is a good reason not to. I don't care about consistency with JS, cause we can't change JS, obv.

satojk: Updated ThingTalk surface syntax for consistency with NN syntax (e.g. `new DatePiece(, , 10, time)`)

lib/date_utils.js
```diff
@@ -88,12 +88,37 @@ function createEdgeDate(edge, unit) {
...
+    if (month > 0) {
+        date.setMonth(month - 1, 1);
+        date.setHours(0, 0, 0, 0);
+    }
+    if (day > 0) {
+        date.setDate(day);
+        date.setHours(0, 0, 0, 0);
+    }
+    if (time !== null)
+        date.setHours(time.hour, time.minute, 0, 0);
```
gcampax, line 110: time.seconds too. And also, time.seconds can be fractional to have millisecond resolution, so don't pass the last `0` to setHours.

lib/describe.js
```diff
@@ -81,11 +81,56 @@ class Describer {
...
+                               "??" :
+                               date.month.toString().padStart(2, '0'));
+            let day = this._(date.day === null ?
+                             "??" :
+                             date.day.toString().padStart(2, '0'));
+            let time = this._(date.time === null ?
+                              "??:??" :
+                              this._describeTime(date.time));
+            base = this._interp(this._("${year}-${month}-${day} ${time}"),
+                                { year, month, day, time });
```
gcampax, line 133: This description is not readable, so it's not suitable. "describe.js" is about deterministic human-readable representation of the code, used to confirm the agent's behavior.

lib/grammar.pegjs
```diff
@@ -810,6 +810,12 @@ time_value = ('makeTime' / 'new' __ 'Time') _ '(' hour:literal_number _ ',' _ mi
 } / '$context' _ '.' _ 'time' _ '.' _ ctx:('morning' / 'evening') {
     return new Ast.Value.Time(new Ast.Time.Relative(ctx));
 }
+piece_date_value = ('makeDatePiece' / 'new' __ 'DatePiece') _ '(' _ year:(literal_number _)? ',' _ month:(literal_number _)? ',' _ day:(literal_number _)? ',' _ time:(time_value _)? ')' {
```
gcampax, line 813: `makeX` is our old syntax, it's deprecated everywhere, we don't need it in new stuff.

Also, if you put `_` outside of the parenthesis and write `year:literal_number? _` you can simplify how you check that the year is present or not, because a missing value will be null, and `_` (whitespace) is always optional.

lib/nn-syntax/tonn_converter.js
```diff
@@ -282,12 +282,37 @@ module.exports = class ToNNConverter {
...
         } else if (value.isDate) {
-            if (value.value === null)
+            if (value.value === null) {
                 return 'now';
-            else if (value.value instanceof Ast.DateEdge)
+            } else if (value.value instanceof Ast.DateEdge) {
                 return List.concat(value.value.edge, 'unit:' + value.value.unit);
-            else
+            } else if (value.value instanceof Ast.DatePiece) {
+                let toConcat = ['new', 'DatePiece', '('];
```
gcampax, line 290: `List` are designed for O(1) concatenation, so you don't need to wrangle with Arrays. You can write:
```
let stuff = List.concat('new', 'DatePiece', '(');
if (value.value.originalYear !== null) {
    stuff = List.concat(stuff, String(value.value.originalYear));
}
```
etc. etc.

lib/nn-syntax/tonn_converter.js
```diff
@@ -282,12 +282,37 @@ module.exports = class ToNNConverter {
...
+                if (value.value.originalYear !== null) {
+                    if (isSmallInteger(value.value.originalYear))
+                        toConcat.push(String(value.value.originalYear));
```
gcampax, line 293: I wonder if we can do something smarter in this code and avoid having to track "originalYear". For example, for all years between 1950 and 2050 we can first check if the sentence has the four digit year (`findEntity(... , { ignoreNotFound: true })`, and then try look for the two digit year or represent as a small number.
Primarily because it would be weird that paraphrasing 2008 to '08 suddenly changed the ThingTalk representation in a way that we cannot automatically convert one to the other.

gcampax:
Also, currently the tokenizer recognizes "year + month" and "month + day" as a full date. What's the plan to address that? Are we changing the tokenizer instead?

satojk:
> , currently the tokenizer recognizes "year + month" and "month + day" as a full date. What's the plan to address that? Are we changing the tokenizer instead?

I think that'd make sense. For month+day I imagine the values shouldn't be dependent on the year the tokenizer is run. For year + month, I think it'd be nice to preserve the information that the utterance did not in fact contain a full date for as long as possible i.e. up until compilation to full date.

Thanks for all the comments! I'll get to work on those.

lib/grammar.pegjs
```diff
@@ -810,6 +816,12 @@ time_value = ('makeTime' / 'new' __ 'Time') _ '(' hour:literal_number _ ',' _ mi
...
+    return new Ast.Value.Date(new Ast.DatePiece(year !== null ? year[0] : null,
+                                                month !== null ? month[0] : null,
+                                                day !== null ? day[0] : null,
+                                                time !== null ? time[0] : null));
+}
```
gcampax, line 824: This is redundant now, right?

lib/nn-syntax/tonn_converter.js
```diff
@@ -282,12 +282,47 @@ module.exports = class ToNNConverter {
...
                 return 'now';
-            else if (value.value instanceof Ast.DateEdge)
+            } else if (value.value instanceof Ast.DateEdge) {
                 return List.concat(value.value.edge, 'unit:' + value.value.unit);
-            else
+            } else if (value.value instanceof Ast.DatePiece) {
+                let toReturn = List.concat('new', 'Date', '(');
+                if (value.value.year !== null) {
+                    let year;
+                    if (value.value.year < 1950) {
```
gcampax, line 293: || year >= 2050

PR #234, opened 2020-08-03
call hooks in compiler

ryachen01:


gcampax:
You should have squashed the commit, to clean up the history (there is a UI option called "squash and merge" in github)
Also, you should wrap your commit messages at around 80 characters (basically, wrap where the instruction text in the editor window wraps)

ryachen01:
Sorry about that I meant to squash and merge, but I accidentally reloaded at some point and didn't realize it was reset to merge.

PR #239, opened 2020-08-06
remove accents for substring matching

ryachen01:
If I search for songs by beyonce in spotify, the request fails because the correct spelling is beyoncé, so the substring match fails on the accent.


ryachen01:
I noticed in the spotify eval set we use hard matching for artists and albums, so should getting rid of accents be expanded to == and contains as well or will that cause problems?

gcampax:
We use `==` for entities, which are compared by ID not by name.

PR #241, opened 2020-08-10
Recurrent time specification & weekdays

rayslxu:


lib/grammar.pegjs
```diff
@@ -787,6 +787,13 @@ short_date_value = ('makeDate' / 'new' __ 'Date') _ '(' _ year:literal_number? _
     d.setMilliseconds(0);
     return new Ast.Value.Date(d);
 }
+long_weekday_value = 'new' __ 'Date' _ '(' _ day:quoted_weekday _ ',' _ hours:literal_number _ ',' _ minutes:literal_number _ ',' _ seconds:literal_number _ ')' {
```
gcampax, line 790: Do you really want the weekday to be quoted? Why?

Also, don't do this: use `ident` or `literal_string` and throw a type error if the day is incorrect. You need to anyway because the NN syntax doesn't check that the day is valid, and that way you don't get an obscure syntax error at the beginning of the whole expression.

lib/nn-syntax/tonn_converter.js
```diff
@@ -319,8 +319,15 @@ module.exports = class ToNNConverter {
                     toReturn = List.concat(toReturn, this.findEntity('TIME', new Ast.Value.Time(value.value.time.value)));
                 toReturn = List.concat(toReturn, ')');
                 return toReturn;
-            }
-            else {
+            } else if (value.value instanceof Ast.WeekDay) {
+                let toReturn = List.concat('new', 'Date', '(', 'enum:' + value.value.weekday);
+                if (value.value.time !== null) {
+                    toReturn = List.concat(toReturn, ',');
+                    toReturn = List.concat(toReturn, this.findEntity('TIME', new Ast.Value.Time(value.value.time.value)));
```
gcampax, line 326: You can concat two things in one statement:
```
toReturn = List.concat(toReturn, ',', this.findEntity('TIME', new Ast.Value.Time(value.value.time.value)));
```

PR #246, opened 2020-08-27
nn-converter: add support for "numeric dates"

gcampax:
The neural network might decide to generate "new Date ( number )"
because that's valid syntax. Later, when we normalize we fail
because there is no DATE token. The syntax generated by the neural
network is incorrect, but we don't want to crash.


gcampax:
This is only necessary to evaluate recent Thingpedia models, so it's not a huge priority.

PR #251, opened 2020-09-07
Improve readability of argument annotations

rayslxu:
The current prettyprint format is really not readable especially when we have
a long list of canonical annotations for arguments, to improve readability:
- linebreak for all argument annotations
- linebreak for first level items of json for argument annotations

### DIFF
--- a/lib/ast/function_def.js
+++ b/lib/ast/function_def.js
@@ -249,7 +249,7 @@ class ArgumentDef extends Node {
      * @return {string} - the ThingTalk code that corresponds to this argument definition
      */
     toString(prefix = '') {
-        return `${this.direction} ${this.name}: ${prettyprintType(this.type, prefix)}${prettyprintAnnotations(this, ' ', false)}`;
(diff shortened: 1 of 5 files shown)


gcampax (commented):
I like this, but how can `#_[formatted]` is not prettyprinted the same way?

rayslxu:
> I like this, but how can `#_[formatted]` is not prettyprinted the same way?

I didn't apply linebreak on arrays to avoid making the output to be extremely long.

rayslxu:
> I like this, but how can `#_[formatted]` is not prettyprinted the same way?

Maybe I can apply it to first level objects in arrays, and things will look like this:
```
#_[formatted=["Here is something for you", {
    type="rdl",
    displayTitle="$title",
    webCallback="$url"
  }]]
```

gcampax:
I see. No I think your end result looks good, no need for changes. I just wanted to understand the implementation.

gcampax (approved):

PR #252, opened 2020-09-09
substring match uses edit distance

ryachen01:
Currently if I say play Gigi D 'agostino spotify will return no results because they have the name as Gigi D'Agostino. We should consider these to be a match because they're just one character apart.

### DIFF
--- a/lib/builtin/primitive_ops.js
+++ b/lib/builtin/primitive_ops.js
@@ -183,12 +183,26 @@ function likeTest(a, b) {
         a = a.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
         b = b.toLowerCase();
         b = b.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
-        if (editDistance(a, b) <= 1 && b.length > 1)
-            return true;
+        a = a.replace(/[^\w\s]|_/g, ' ');
+        a = a.replace(/ +(?= )/g, '');
+        a = a.trim();
+        b = b.replace(/[^\w\s]|_/g, ' ');
+        b = b.replace(/ +(?= )/g, '');
+        b = b.trim();
         if (a.indexOf(b) >= 0)
             return true;
-        if (a.replace(/'/g, ' ').indexOf(b.replace(/'/g, ' ')) >= 0)
-            return true;
+        for (let token_b of b.split(' ')) {
+            let tokenFound = false;
+            for (let token_a of a.split(' ')) {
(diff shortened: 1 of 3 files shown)


gcampax (commented):
Is "1" the right threshold for edit distance?
We consider substrings to be a match regardless of the number of unmatched characters, so what if there is a substring that is one-off from the string to search?

ryachen01:
function fuzzyMatch(pattern, str) {
      pattern = '.*' + pattern.split('').join('.*') + '.*';
      const re = new RegExp(pattern);
      return re.test(str);
}
What if we did something like this. This will return true as long as all of the letters of the search are in the right order in the matching result.

gcampax:
That's probably too lose. Actually, going back to the problem, I'm more inclined to say we strip out all accents, replace apostrophes and other punctutation with spaces, normalize the spaces, and then do a substring match. Cause really we want to match words with words. We can even do a token by token edit-distance match, so two tokens are considered to match if the edit distance is at most 1 (so we catch singular/plural and typos).

lib/builtin/primitive_ops.js
```diff
@@ -143,10 +183,26 @@ function likeTest(a, b) {
         a = a.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
         b = b.toLowerCase();
         b = b.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
+        a = a.replace(/[^\w\s]|_/g, ' ');
```
gcampax, line 186: This pattern will replace all non-ASCII characters, including CJK characters, Arabic characters, etc. We probably don't want that.
A better regex might be `/[\p{Mark}\p{Punctuation}\p{Separator}\p{Other}_]/u` (note the u at the end)

lib/builtin/primitive_ops.js
```diff
@@ -143,10 +183,26 @@ function likeTest(a, b) {
...
+        a = a.replace(/ +(?= )/g, '');
```
gcampax, line 187:
```suggestion
        a = a.replace(/\p{White_Space}+/ug, ' ');
```

gcampax:
We should also write tests for this change. A lot of tests.

gcampax:
Mh looks like eslint fails. I think you need to change .eslintrc.yml, set ecmaVersion: 2018 because we're using the new Unicode escapes in the regular expressions.

gcampax (approved):
Looks good!

PR #254, opened 2020-09-14
Update Wikidata SPARQL converter

rayslxu:
Update the converter to match the new ThingTalk syntax we added in the past year. 
The old code was also problematic. Most of the issues are fixed. Or operator is still broken, we will fix it later.

### DIFF
--- a/lib/wikidata_sparql.js
+++ b/lib/wikidata_sparql.js
@@ -54,12 +54,11 @@ module.exports = class WikidataSparqlConverter {
 
     _getSparqlVar(scope, name) {
         let scopeEntry;
-        if (name.isVarRef) {
+        if (name.isVarRef)
             scopeEntry = scope[name.name];
-            name = name.name.split(".")[1];
-        } else {
+        else
             scopeEntry = scope[name];
-        }
+
 
         if (scopeEntry.isId) {
             return "?table" + scopeEntry.tableIndex;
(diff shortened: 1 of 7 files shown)


gcampax (commented):
I think this looks good.
I'm not sure about the part where you say "We don't know if a value is represented as a string or an array". Wouldn't we know that from the type?

rayslxu:
> I think this looks good.
> I'm not sure about the part where you say "We don't know if a value is represented as a string or an array". Wouldn't we know that from the type?

do you mean this one: "we don't know if a value is represented as a string or an entity in Wikidata"
Lots of things we considered String is also an entity in wikidata (like given name, last name), so we cannot decide based on thingtalk type.

gcampax:
Yes I meant string or entity (brain fart, probably).
Time to add a new annotation with the wikidata type? We did that for schema.org (even though we didn't really use it)

gcampax:
Mostly I ask because it looks like you're generating some really inefficient SPARQL and I worry things will timeout more often than not.

rayslxu:
> Mostly I ask because it looks like you're generating some really inefficient SPARQL and I worry things will timeout more often than not.

I tested that, in fact there is no significant difference in terms of time. but yeah, we should probably include type information - it's going to be tricky though because the same property can be used for different types of values (even for the same domain). Anyway, it's going to be part of the project to dump the entire wikdiata.

PR #258, opened 2020-09-16
More SPARQL conversion fixes

rayslxu:


lib/wikidata_sparql.js
```diff
@@ -299,6 +339,9 @@ module.exports = class WikidataSparqlConverter {
             let filterRHS;
             let filterLHS;
 
+            if (value.isLocation && (!value.lat || !value.lon))
+                throw new Error('Conversion to SPARQL for location type is not supported yet.');
```
gcampax, line 343: shouldn't it be value.value.lat ?

gcampax (approved):
Yeah this looks good.

PR #268, opened 2020-09-30
Third TypeScript PR

gcampax:
This is going to be the last of the big TypeScript conversions. After this, all that's left is syntax (grammar, prettyprint, nn-syntax), describe, and the permission control stuff. I'm not sure about the future of Describe, and I definitely don't want to touch the permission control stuff, so that leaves the syntax stuff, which is going to be compat-only as soon as we have the new syntax.


rayslxu (approved):
lgtm (my questions have been answered offline)

PR #273, opened 2020-10-13
The beginning of ThingTalk 2.0

gcampax:
The new syntax, and the first round of AST changes.

This is on top of #270


api-changes.md
```diff
@@ -3,6 +3,120 @@
...
+  ```
+  [distance(geo, $location.current_location) as dist] of @com.yelp.restaurant()
+  ```
+
+- `join` is now expressed as `=>` (true table joins should be expressed as
+  subqueries instead)
+  
+- `monitor` of a specific field is `monitor(fields of table)`; example:
+  ```
+  monitor(text, author of @com.twitter.home_timeline())
```
rayslxu, line 63: It's a little inconsistent here between monitor and projection. could we remove the `[]` in projection? otherwise, we should probably add `[]` here.

api-changes.md
```diff
@@ -3,6 +3,120 @@
...
 into HISTORY when the release is cut.
 
+## AST changes
+
 - Remove non-device selectors.
 
+- Table, Stream, Actions are no longer used. Instead, everything is
+  an Expression, which could be a table Expression (Index, Sort, Aggregation),
+  a table-or-stream Expression (Filter, Projection), or any of the
+  three types (FunctionCall, Invocation)
```
rayslxu, line 13: what three types? what is the difference between `FunctionCall` and `Invocation`?

gcampax:
@rayslxu this is the full ThingTalk 2.0 PR now: it includes the full ThingTalk functionality in the new syntax, ports over all the tests, and makes the new syntax the default. Parsing and serialization goes through `Syntax.parse` and `Syntax.serialize`, which replace `Grammar.parse` and take as input a `SyntaxType` to determine which way it goes: new syntax, new syntax in tokenized form, legacy syntax, or legacy nn.

The new AST nodes have a `toLegacy()` method to convert to the old Stream/Table/Action classes (_after_ typechecking). A number of modules in ThingTalk rely on `toLegacy()` because they have not ported over to the new AST objects, including:
- the compiler
- describe.js
- permission checker
- wikidata sparql

Conversely, the legacy classes have a `toExpression()` method to convert to the new classes. This is necessary because the top level Ast objects like Ast.Program need the new classes.

rayslxu:
Before I dive deeper, I think we need boolean question representation - it's required by CSQA if we want to have it in the paper. Probably in the form `any(<table>)` ?

gcampax:
I think boolean questions should be framed as computations that return a boolean...

Btw, on the synthesis we'll need an option to not represent boolean questions that way, because we already have a lot of boolean questions in multiwoz and we just represent them as projections. I don't know if we want to correct that aspect because it might involve manual reannotation.

tools/generate-parser/index.ts
```diff
@@ -128,7 +129,11 @@ async function main() {
         }
     }
 
-    const generator = new SLRParserGenerator(grammar, 'input', typeMap['input'] || 'any');
-    await writeout(firstFile.preamble, generator, fs.createWriteStream(output), output, typeMap['input'] || 'any');
+    if (output === '--wsn') {
+        wsnDump(grammar);
```
rayslxu, line 133: what does `wsn` stand for?

lib/typecheck.ts
```diff
@@ -392,1195 +421,1293 @@ async function typeCheckFilter(ast : Ast.BooleanExpression,
...
+            this._addRequiredInputParamsInvocation(action, new Set<string>());
+        else if (action instanceof Ast.InvocationAction)
+            this._addRequiredInputParamsInvocation(action.invocation, new Set<string>());
+        else
+            throw new TypeError();
     }
-}
 
-function typeCheckMetadata(func : unknown) {
-    /*Object.entries(func.metadata).forEach(([name, value]) => {
```
rayslxu, line 1067: what happened to this?

lib/typecheck.ts
```diff
@@ -392,1195 +421,1293 @@ async function typeCheckFilter(ast : Ast.BooleanExpression,
...
+        });
 
-    // remove all input parameters (which we have filled with $undefined)
-    const args = Array.from(ast.value.schema!.iterateArguments()).filter((a) => !a.is_input);
-    ast.schema = new Ast.FunctionDef(ast.location, 'query', null, ast.name, [],
-        { is_list: ast.value.schema!.is_list, is_monitorable: false }, args, {});
-    scope.addGlobal(ast.name, ast.schema);
-}
+        this._typeCheckMetadata(ast);
+        this._typeCheckProgramAnnotations(ast);
```
rayslxu, line 1218: `this._typeCheckProgramAnnotations` should be `async` and we should need `await` here?

lib/typecheck.ts
```diff
@@ -392,1195 +421,1293 @@ async function typeCheckFilter(ast : Ast.BooleanExpression,
...
+            else
+                await this._typeCheckExpressionStatement(decl, scope);
+        }
     }
-    for (const choice of ast.oninputs) {
-        scope.clean();
-        await typeCheckRule(choice, schemas, scope, classes, useMeta);
+
+    async typeCheckProgram(ast : Ast.Program,
+                           parentScope : Scope|null = null) : Promise<void> {
```
rayslxu: why is this function of type `Promise<void>`?  there are a few other cases where nothing is returned, but the function is annotated with `Promise<void>`; there are also a few other cases that are the other way around.

rayslxu (commented):
Overall looks good. 

For boolean questions, I saw you have the syntax with filter expression in projection. Could you also have a few examples in test for that (I use those examples as a quick way to check syntax from time to time)

tools/generate-parser/index.ts
gcampax: https://en.wikipedia.org/wiki/Wirth_syntax_notation

It's an alternative to BNF, and it's one of the input formats of [kgt](https://github.com/katef/kgt), the tool I found to generate the documentation for the grammar.

lib/typecheck.ts
gcampax: This code has been commented out fro a while, because it doesn't work with the latest format of canonical forms (with enums, argmin/max, etc.)

lib/typecheck.ts
gcampax: Async functions that return nothing return `Promise<void>`, and regular functions that return nothing are `void`. In both cases the return type annotation can usually be omitted and typescript infers it, except for publicly exported functions and public methods of exported classes we used to have a eslint warning if the function lacks a type annotation. I recently silenced it because I got tired of typing, but yeah.

PR #280, opened 2020-12-05
Another round of fixes to get Genie going

gcampax:
Turns out Genie needs a few more fixes here before its own tests pass.


gcampax:
I'm going to merge this to keep things going, we can adjust later.

PR #282, opened 2020-12-08
Fix makefile target dependency

nschuc:
Build only completes successfully on the second execution. This prevents installing `genie-toolkit#next` as well.

Not a Make expert so there may be a better way to address this dependency problem.


gcampax:
Yep, this looks correct. Thanks for catching this!

PR #283, opened 2020-12-09
Make empty parenthesis optional in Thingpedia function calls

gcampax:
The old NN syntax did not have parenthesis around parameters at
all, so now the semantic parser is forced to predict a pair of
parenthesis, which is redundant in the common case of KBQA (queries
as tables, no input parameters). This commit makes it so that
empty parenthesis can be omitted altogether.


gcampax:
Parenthesis don't seem to be problematic after all.

PR #288, opened 2020-12-29
Remove all code dealing with natural language

gcampax:
I want to move the Formatter and Describe modules to Genie. This will consolidate all code that deals with natural language in Genie, and let me integrate those parts more closely with the dialogue state machine.
In particular, my goal is to use the Formatter only for `$result` (formerly `$event`) and for card outputs (pictures, RDLs), which will not go through the dialogue state machine and the speech output. My other goal is to use Describe to get confirmation strings in the dialogue state machine in an efficient manner, which I need to do streams and compound commands. As a side goal, I plan to kill `#_[confirmation]` and use a combination of `#_[canonical]` and `dataset` exclusively.


rayslxu (approved):
lgtm

PR #294, opened 2021-01-27
[AL1-14] Implement entity inheritance

gcampax:
Entity inheritance (subtyping) is the ability to declare that a certain custom entity type is assignable to another entity type.
The syntax is the obvious one: `entity foo extends bar;`

This feature is needed for MultiWOZ to convert everything from strings to entities (which fixes some issues in generating cross-domain coreferences and fixes requoting MultiWOZ datasets), because the taxi action needs to take in a place that can be a restaurant name, an hotel name, or an attraction name.
It is also needed for Spotify, because we want song and album to be both assignable to a parameter of play().

Entity inheritance combines nicely with function inheritance: if two queries inherit from each other, they should also declare the type of their `id` parameters as subtypes.

### DIFF
--- a/lib/ast/program.ts
+++ b/lib/ast/program.ts
@@ -1261,6 +1261,7 @@ export class MixinImportStmt extends Node {
 export class EntityDef extends Node {
     isEntityDef = true;
     name : string;
+    extends : string|null;
     nl_annotations : NLAnnotationMap;
     impl_annotations : AnnotationMap;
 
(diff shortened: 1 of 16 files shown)


gcampax:
@AydanPirani @ryachen01 this PR will be needed to complete the Spotify cleanup when we get to the generation parts.

rayslxu (approved):
lgtm. This is also useful for wikidata project. It's probably good for AutoQA in general: we should take into subtyping into consideration when choosing example values. E.g., given universities are subtype of organization, we can avoid using universities as example values for `worksFor`, so we don't get confused with `alumniOf`.

PR #296, opened 2021-01-28
[AL1-3] Basic subquery support 

rayslxu:
Add support for comparison subquery. No parameter passing support in this version, which will be added in the next beta version.

### DIFF
--- a/lib/ast/expression.ts
+++ b/lib/ast/expression.ts
@@ -24,6 +24,7 @@ import Node, { SourceRange } from './base';
 import NodeVisitor from './visitor';
 import { FunctionDef } from './function_def';
 import { Value } from './values';
+import { Expression } from './expression2';
 
 import Type from '../type';
 import * as Optimizer from '../optimize';
@@ -391,6 +392,7 @@ export class Invocation extends Node {
  * @property {boolean} isAtom - true if this is an instance of {@link Ast.BooleanExpression.Atom}
  * @property {boolean} isNot - true if this is an instance of {@link Ast.BooleanExpression.Not}
  * @property {boolean} isExternal - true if this is an instance of {@link Ast.BooleanExpression.External}
+ * @property {boolean} isComparisonSubquery - true if is an instance of {@link Ast.BooleanExpression.ComparisonSubquery}
  * @property {boolean} isTrue - true if this is {@link Ast.BooleanExpression.True}
  * @property {boolean} isFalse - true if this is {@link Ast.BooleanExpression.False}
  * @property {boolean} isCompute - true if this is {@link Ast.BooleanExpression.Compute}
@@ -406,6 +408,8 @@ export abstract class BooleanExpression extends Node {
     isNot ! : boolean;
     static External : any;
     isExternal ! : boolean;
+    static ComparisonSubquery : any;
+    isComparisonSubquery ! : boolean;
     static True : any;
     isTrue ! : boolean;
     static False : any;
@@ -445,6 +449,7 @@ BooleanExpression.prototype.isOr = false;
 BooleanExpression.prototype.isAtom = false;
 BooleanExpression.prototype.isNot = false;
 BooleanExpression.prototype.isExternal = false;
+BooleanExpression.prototype.isComparisonSubquery = false;
 BooleanExpression.prototype.isTrue = false;
 BooleanExpression.prototype.isFalse = false;
 BooleanExpression.prototype.isCompute = false;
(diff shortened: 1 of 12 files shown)


lib/ast/expression.ts
```diff
@@ -919,6 +924,95 @@ export class ExternalBooleanExpression extends BooleanExpression {
...
+    *iterateSlots(schema : FunctionDef|null,
+                  prim : InvocationLike|null,
+                  scope : ScopeMap) : Generator<OldSlot, void> {
+        yield* this.rhs.iterateSlots(scope);
+    }
+
+    *iterateSlots2(schema : FunctionDef|null,
+                   prim : InvocationLike|null,
+                   scope : ScopeMap) : Generator<DeviceSelector|AbstractSlot, void> {
+        yield* this.rhs.iterateSlots2(scope);
```
gcampax, line 1010: lhs would be a slot too right?

lib/ast/expression2.ts
```diff
@@ -74,6 +88,7 @@ export abstract class Expression extends Node {
     abstract toLegacy(into_params ?: InputParam[], scope_params ?: string[]) : legacy.Stream|legacy.Table|legacy.Action;
     abstract clone() : Expression;
     abstract toSource() : TokenStream;
+    abstract equals(other : Expression) : boolean;
```
gcampax, line 91: This is quite useful, because it means we can clean up the equals implementation in DialogueStateHistoryItem

lib/compiler/ast-to-ops.ts
```diff
@@ -730,8 +737,74 @@ function compileStatementToOp(statement : Ast.Rule|Ast.Command) : RuleOp {
     return new RuleOp(streamop, statement.actions, statement);
 }
 
+function compileBooleanExpressionToOp(expr : Ast.BooleanExpression) : BooleanExpressionOp {
+    if (expr instanceof Ast.AtomBooleanExpression)
```
gcampax, line 741: If we're going the route of an intermediate representation for boolean expressions, we should probably remove the distinction of AtomBooleanExpression and ComputeBooleanExpression at the "op" level.

lib/compiler/ast-to-ops.ts
```diff
@@ -730,8 +737,74 @@ function compileStatementToOp(statement : Ast.Rule|Ast.Command) : RuleOp {
...
+            expr.channel,
+            expr.in_params,
+            compileBooleanExpressionToOp(expr.filter),
+            expr.schema,
+            expr.__effectiveSelector
+        );
+    }
+
+    if (expr instanceof Ast.ComparisonSubqueryBooleanExpression) {
+        assert(expr.rhs instanceof Ast.ProjectionExpression && expr.rhs.args.length === 1);
```
gcampax, line 774: can rhs have a computation instead of a parameter projection?

lib/compiler/ast-to-ops.ts
```diff
@@ -730,8 +737,74 @@ function compileStatementToOp(statement : Ast.Rule|Ast.Command) : RuleOp {
...
+        return BooleanExpressionOp.True;
+
+    if (expr === Ast.BooleanExpression.False)
+        return BooleanExpressionOp.False;
+
+    if (expr instanceof Ast.ComputeBooleanExpression)
+        return new BooleanExpressionOp.Compute(expr, expr.lhs, expr.operator, expr.rhs, expr.overload);
+
+    if (expr instanceof Ast.DontCareBooleanExpression)
+        return new BooleanExpressionOp.DontCare(expr, expr.name);
```
gcampax, line 799: As far as the compiler is concerned, DontCare is the same as True

gcampax (commented):
This looks good, and seems straightforward.

Do we need to change optimize.ts to run the normalization inside the subquery as well?

lib/compiler/ast-to-ops.ts
rayslxu: why do we want to do that? these two ops are compiled differently, we will need to distinguish them by fields anyway if we merge them.

gcampax: Because you don't need to compile them differently: an AtomBooleanExpression is the same as a ComputeBooleanExpression where the lhs is a VarRefValue.

rayslxu:
> This looks good, and seems straightforward.
> 
> Do we need to change optimize.ts to run the normalization inside the subquery as well?

The optimize.ts has been changed already - it is just one line. 

I have simplified the BooleanExpressionOp as you suggested; computation projection also works now; iterateSlots2 is fixed, while iterateSlot cannot be supported.

gcampax (approved):
I think this has all the pieces! Maybe more test cases (eg test_nn_syntax, test_iteration_apis, test_legacy_syntax, etc.) but those can come later.

PR #310, opened 2021-02-24
Add support for existential subquery

rayslxu:
Existential subquery generalizes the old get predicate (`ExternalBooleanExpression`) as it has full support for all possible queries, such as projection, sort, computation. 

`ExternalBooleanExpression` is not removed for compatibility reasons.

### DIFF
--- a/lib/compiler/ops-to-jsir.ts
+++ b/lib/compiler/ops-to-jsir.ts
@@ -306,6 +306,20 @@ export default class OpCompiler {
 
             this._irBuilder.popBlock(); // for-of
             this._irBuilder.popBlock(); // try-catch
+        } else if (expr instanceof Ops.ExistentialSubqueryBooleanExpressionOp) {
+            this._irBuilder.add(new JSIr.LoadConstant(new Ast.Value.Boolean(false), cond));
+
+            const blockStack = this._irBuilder.saveStackState();
+            const tmpScope = this._currentScope;
+
+            this._compileTable(expr.subquery);
+
+            this._irBuilder.add(new JSIr.LoadConstant(new Ast.Value.Boolean(true), cond));
+            this._irBuilder.add(new JSIr.Break());
+            this._irBuilder.popBlock();
+
+            this._currentScope = tmpScope;
+            this._irBuilder.popTo(blockStack);
         } else if (expr instanceof Ops.ComparisonSubqueryBooleanExpressionOp) {
             this._irBuilder.add(new JSIr.LoadConstant(new Ast.Value.Boolean(false), cond));
 
--- a/lib/typecheck.ts
+++ b/lib/typecheck.ts
@@ -473,13 +473,25 @@ export default class TypeChecker {
             return;
         }
 
-        assert(ast instanceof Ast.ExternalBooleanExpression);
-        if (ast.schema === null)
-            await this._loadTpSchema(ast);
-        if (ast.schema!.functionType !== 'query')
-            throw new TypeError(`Subquery function must be a query, not ${ast.schema!.functionType}`);
-        await this._typeCheckInputArgs(ast, ast.schema!, scope);
-        await this._typeCheckFilterHelper(ast.filter, ast.schema, scope);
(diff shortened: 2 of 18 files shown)


gcampax (approved):
This looks good I think. Maybe add more syntax and typechecking tests?
Especially for serializing to legacy syntax and legacy NN syntax, which is needed when the NLP server talks to older clients.

Also, by 2.0 stable we should get rid of all the "compatibility" AST nodes. We can keep the syntax but we don't need legacy internal interfaces.

rayslxu:
> This looks good I think. Maybe add more syntax and typechecking tests?
> Especially for serializing to legacy syntax and legacy NN syntax, which is needed when the NLP server talks to older clients.
> 
> Also, by 2.0 stable we should get rid of all the "compatibility" AST nodes. We can keep the syntax but we don't need legacy internal interfaces.

Just realize that I need to convert some of the get predicates into comparison subquery - when there is a filter on parameters of the main query. I will change this to a draft for now until I fix that.

test/test_syntax.tt
```diff
@@ -1887,4 +1887,37 @@ now => @com.facebook.post();
...
+// existential subquery
+// simple invocation no filter
+monitor(@com.twitter.home_timeline() filter any(@org.thingpedia.builtin.thingengine.builtin.get_gps()));
+
+
+====
+
+// existential subquery
+// projection
+monitor(@com.twitter.home_timeline() filter any([speed] of @org.thingpedia.builtin.thingengine.builtin.get_gps()));
```
gcampax, line 1903: Should we optimize out projections inside existential subqueries? They have no effect, right?

lib/typecheck.ts
```diff
@@ -473,13 +473,25 @@ export default class TypeChecker {
...
+
+        assert(ast instanceof Ast.ExistentialSubqueryBooleanExpression);
+        await this._typeCheckSubquery(ast.subquery, scope);
+    }
+
+    private async _typeCheckSubquery(expr : Ast.Expression, scope : Scope) {
+        if (expr.schema === null)
+            await this._loadAllSchemas(expr);
+        await this._typeCheckExpression(expr, scope);
+        this._checkExpressionType(expr, ['query'], 'projection');
```
gcampax, line 494:
```suggestion
        this._checkExpressionType(expr, ['query'], 'subquery');
```

lib/compiler/ops-to-jsir.ts
```diff
@@ -306,6 +306,20 @@ export default class OpCompiler {
...
+            this._irBuilder.add(new JSIr.LoadConstant(new Ast.Value.Boolean(false), cond));
+
+            const blockStack = this._irBuilder.saveStackState();
+            const tmpScope = this._currentScope;
+
+            this._compileTable(expr.subquery);
+
+            this._irBuilder.add(new JSIr.LoadConstant(new Ast.Value.Boolean(true), cond));
+            this._irBuilder.add(new JSIr.Break());
+            this._irBuilder.popBlock();
```
gcampax, line 319: I'm not exactly sure this code works if the expression inside requires a reduction (eg aggregation, index, argmin/argmax). I think we need a test.

I'm also not exactly sure what the semantics of those would be inside an existential subquery. Maybe they need to be optimized out? Or maybe only index is preserved, and the semantics mean "at least this many rows were returned"?

gcampax (approved):
I think this looks good, but I have a few suggestions.

test/test_syntax.tt
rayslxu: I think it's possible for a query to have an optional output parameter (such as get tweet may contain a picture or not). As we group functionalities based on natural language, we probably will keep this design. So projection here is meaningful. I think the compiler code is not reflecting this design though... so that needs to be updated...

gcampax: ok so the semantics of existential projection means "there exist at least one row for which all the projected values are not `null` or `undefined`" is that right?

rayslxu: right

PR #318, opened 2021-03-03
Fix inadvertently merging location entities

gcampax:
Two locations that are less than 0.1 degree apart in either coordinate
can be quite far (for latitude, 0.1 degrees is about 7 miles).
Instead, we should compare a lot more strictly, so we only catch
floating point rounding errors (due to roundtripping from JSON/strings
in various places) and not locations that are actually distinct.

### DIFF
--- a/lib/entity-retriever.ts
+++ b/lib/entity-retriever.ts
@@ -35,6 +35,8 @@ import {
     EntityMap
 } from './entities';
 
+const EPSILON = 1e-8;
+
 function entitiesEqual(type : string, one : AnyEntity, two : AnyEntity) : boolean {
     if (one === two)
         return true;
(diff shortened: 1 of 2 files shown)


rayslxu (commented):
Will this affect comparison between current location and a location entity?

gcampax:
`==` for locations (aka `locationEqual`) is intentionally loose because otherwise "restaurants in Palo Alto" gives you only a tiny radius, and it's even worse for "restaurants in San Francisco". We can reduce the radius to 1km perhaps.
(Long term, we should have a new design for locations that accounts for radius and/or administrative boundaries given the precision of the user's specification, so you can also ask for "restaurants on University Avenue" or "restaurants in Santa Clara County". I'm not sure when that will happen)

rayslxu:
> `==` for locations (aka `locationEqual`) is intentionally loose because otherwise "restaurants in Palo Alto" gives you only a tiny radius, and it's even worse for "restaurants in San Francisco". We can reduce the radius to 1km perhaps.
> (Long term, we should have a new design for locations that accounts for radius and/or administrative boundaries given the precision of the user's specification, so you can also ask for "restaurants on University Avenue" or "restaurants in Santa Clara County". I'm not sure when that will happen)

I see, let's keep it as 10km then. We should create an issue for this, once we do rules for home assistant, this is essential

rayslxu (approved):

PR #347, opened 2021-05-16
[WIP] Timers & Reminders

emilyjchang:


lib/ast/primitive.ts
```diff
@@ -971,16 +971,21 @@ Stream.Timer.prototype.isTimer = true;
 export class AtTimerStream extends Stream {
     time : Value[];
     expiration_date : Value|null;
+    once : Value|false;
```
gcampax, line 974: Wait this type doesn't make sense. Ast.Value is not a boolean, it's an object, so the default value should be `new Ast.BooleanValue(false)`

lib/nn-syntax/parser.lr
```diff
@@ -269,8 +269,8 @@ table_join : Ast.JoinTable = {
 stream : Ast.Stream = {
     'timer' 'base' '=' base:value ',' 'interval' '=' interval:value => new Ast.Stream.Timer($.location, base, interval, null, null);
     'timer' 'base' '=' base:value ',' 'interval' '=' interval:value ',' 'frequency' '=' frequency:value => new Ast.Stream.Timer($.location, base, interval, frequency, null);
-    'attimer' 'time' '=' time:value => new Ast.Stream.AtTimer($.location, time instanceof Ast.ArrayValue ? time.value : [time], null, null);
-    'attimer' 'time' '=' time:value ',' 'expiration_date' '=' expiration_date:value => new Ast.Stream.AtTimer($.location, time instanceof Ast.ArrayValue ? time.value : [time], expiration_date, null);
+    'attimer' 'time' '=' time:value => new Ast.Stream.AtTimer($.location, time instanceof Ast.ArrayValue ? time.value : [time], null, false, null);
+    'attimer' 'time' '=' time:value ',' 'expiration_date' '=' expiration_date:value ',' 'once' '=' once:value => new Ast.Stream.AtTimer($.location, time instanceof Ast.ArrayValue ? time.value : [time], expiration_date, once, null);
```
gcampax, line 273: This change is not correct because it makes the `once` parameter required if `expiration_date` is specified.
TBQH I wouldn't touch the legacy syntax because this is a new feature, but if we do add the syntax, we should do it in a compatible way.

lib/ast/primitive.ts
```diff
@@ -971,16 +971,21 @@ Stream.Timer.prototype.isTimer = true;
...
                 time : Value[],
                 expiration_date : Value|null,
+                once : Value|false,
                 schema : FunctionDef|null) {
         super(location, schema);
 
         assert(Array.isArray(time));
         this.time = time;
 
+        assert(once instanceof Value);
```
gcampax, line 986: I'm surprised this assert wasn't a problem given you pass `false` in a bunch of places.

gcampax (commented):
We should add a couple of tests for this:
- adding a new example to test/test_syntax.tt should be enough to cover thingtalk 2.0 syntax and typechecking
- we should add a compiler test in test/test_compiler.js
- given you're adding legacy syntax for this, you should add a test in test/legacy_nn_syntax.js

lib/nn-syntax/parser.lr
emilyjchang: Is the parser.lr just legacy? If so, I can just remove this then.

gcampax: lib/nn-syntax/parser.lr is legacy, lib/new-syntax/parser.lr is current - but the latter should need no changes because timers are treated like regular function calls, and which parameters are allowed is solved by typechecking not syntax.

lib/ast/primitive.ts
emilyjchang: Would it be fine to just use the `boolean` type then? Or is it preferred to use the object

gcampax: It's preferred to use the object. If you use boolean, you need to enforce that the Ast.Value is a constant somewhere, or you'll crash when you call `once.value.toJS()` in `FunctionCallExpression.toLegacy`. Ast.Value is the base class of all scalar expressions in ThingTalk, which can be constants or can be computed expressions. In the syntax, a function can be called with arbitrary values (provided the types match) so you can't statically convert to a boolean. Similarly in the compiler you need to call compileValue so the compiler generates the appropriate JS code for the value.

lib/ast/expression2.ts
```diff
@@ -171,13 +172,15 @@ export class FunctionCallExpression extends Expression {
             } else if (this.name === 'attimer') {
                 const time = this.in_params.find((ip) => ip.name === 'time')!;
                 const expiration_date = this.in_params.find((ip) => ip.name === 'expiration_date');
+                const once = new BooleanValue(false);
```
gcampax, line 175: This is not correct. You should look for the parameter called "once" in the list of parameters to the "attimer" function (aka `this.in_params`)

test/test_compiler.js
```diff
@@ -10409,6 +10409,42 @@ p_recip();`,
...
+  let _t_2;
+  let _t_3;
+  let _t_4;
+  let _t_5;
+  let _t_6;
+  try {
+    _t_1 = new Array(1);
+    _t_2 = new __builtin.Time(12, 30, 0);
+    _t_1[0] = _t_2;
+    _t_0 = await __env.invokeAtTimer(_t_1, null);
```
gcampax, line 10428: Yeah see: there is nothing in this call to invokeAtTimer that passes the once=true, because "once=true" got lost in the translation from Ast.Expression to Ast.Table

lib/compiler/ops.ts
```diff
@@ -134,6 +134,7 @@ StreamOp.Timer = TimerStreamOp;
 export class AtTimerStreamOp extends StreamOp {
     constructor(public time : Ast.Value[],
                 public expiration_date : Ast.Value|null,
+                public once : Ast.Value|null,
                 public ast : Ast.Stream) {
```
gcampax, line 138: You will also need to add code in ops-to-jsir to make use of this field, compiling it to a call to ExecEnvironment.invokeAtTimer with the right parameters.

gcampax (changes requested):

lib/compiler/ops.ts
emilyjchang: I see, just added it

lib/compiler/ops-to-jsir.ts
```diff
@@ -493,9 +493,12 @@ export default class OpCompiler {
         }
 
         let expiration_date = null;
+        let once = null;
         if (streamop.expiration_date !== null)
             expiration_date = this.compileValue(streamop.expiration_date, this._currentScope);
-        this._irBuilder.add(new JSIr.InvokeAtTimer(iterator, timeArray, expiration_date));
+        if (streamop.once !== null)
+            once = this.compileValue(streamop.once, this._currentScope);
```
gcampax, line 500: It's probably cleaner if we handle the case of streamop.once === null here and make it:
```
once = this._irBuilder.allocateRegister();
this._irBuilder.add(new JSIr.LoadConstant(once, false));
```

That way, invokeAtTimer doesn't need to handle a "null" case. Also, we should change the declaration of invokeAtTimer in ExecEnvironment accordingly.

test/test_syntax.tt
```diff
@@ -1016,6 +1016,14 @@ attimer(time=[new Time(9, 0), new Time(15, 0)]) => @org.thingpedia.builtin.thing
 attimer(time=[new Time(9, 0)]) => @org.thingpedia.builtin.thingengine.builtin.say(message="it's 9am or 3pm");
 ====
 
+// ontimer with List
+attimer(time=[new Date(2021, 9, 30), new Date(2021, 9, 30, 12, 30, 0)]) => @org.thingpedia.builtin.thingengine.builtin.say(message="hello");
```
gcampax, line 1020: Typo? Should be `ontimer`

gcampax (approved):
Looks good!
Please change the commit message a bit: format it so it's a short subject line (which should fit in the Github UI without wrappning), an empty line, and the rest of the explanation.

Also, please add the declaration of invokeOnTimer to ExecEnvironment, for documentation and type safety in the implementation.

PR #350, opened 2021-05-30
Two bug fixes to support Spotify currently playing

gcampax:
Needed for https://github.com/stanford-oval/thingpedia-common-devices/pull/343

The first one is an actual bug that could occur with other skills too, the second one relaxes the rules around subfunctions. Subfunctions are basically just a shorthand to declare a bunch of functions with the same output parameters, so we shouldn't impose any monitorability restrictions.


rayslxu (approved):
looks good

PR #354, opened 2021-06-04
Add history query

cffan:


lib/ast/expression.ts
```diff
@@ -279,6 +280,65 @@ export class InvocationExpression extends Expression {
     }
 }
 
+export class HistoryQueryExpression extends Expression {
+    selector : DeviceSelector;
+    channel : string;
+
+    constructor(location : SourceRange|null,
+                selector : DeviceSelector,
```
gcampax, line 288: Given you cannot pass an arbitrary device selector, but only a device kind, I would make HistoryQueryExpression store a `kind` property and avoid the DeviceSelector object.

lib/typecheck.ts
```diff
@@ -411,6 +411,15 @@ export default class TypeChecker {
         return overload;
     }
 
+    private async _typeCheckHistoryQuery(ast : Ast.HistoryQueryExpression) {
+        const classdef = this._classes[ast.selector.kind];
+        assert(classdef);
```
gcampax, line 416: This assertion looks wrong. You will get a classdef here only if the class is defined in the same program as the statement (which is almost never the case - the class is typically defined in thingpedia)

gcampax (commented):
I think this looks ok. You might want to add a compiler and runtime test to be sure.

Another thing I'm wondering about, rather than removing input arguments, we might want to turn them into output arguments, because we'll definitely want to search for history records where we passed certain inputs to a function.

lib/typecheck.ts
cffan: How do I check if the classdef exists?

lib/ast/expression.ts
cffan: Done

gcampax (approved):
Shall we merge this while we work on the Genie side?

PR #356, opened 2021-06-19
Add set_time Scalar Operator

emilyjchang:


lib/ast/values.ts
```diff
@@ -1343,21 +1343,37 @@ function dateToSource(date : DateLike) : TokenStream {
...
+
+        if (time && time.value instanceof AbsoluteTime) {
+            if (value instanceof Date) {
+                value.setHours(time.value.hour);
+                value.setMinutes(time.value.minute);
+                value.setSeconds(time.value.second);
+            } else if (value instanceof WeekDayDate) {
+                if (value.time) 
+                    value = new WeekDayDate(value.weekday, time.value);
+            }
```
gcampax, line 1376: I think it's missing the case where value is a DatePiece but the year is null, in which case we need to overwrite the `time` field of DatePiece with the provided time.

It's also missing the case of a DateEdge ("today", "tomorrow", "this week", etc.).
That needs to be treated like a relative time I think - we'll handle that later. But let's add an assert that it won't happen for now.

gcampax (commented):
We need to add syntax for this as well, I think: we need to be able to parse:
```
new Date ( DATE_0 , TIME_0 )
```
and we need to serialize to that representation if the sentence contains a DATE_0 and a TIME_0.

We also need to add tests for nn-syntax with different combinations of entities recognized by the tokenizer in the sentence.

In fact, I'm almost tempted to say that the current approach is wrong, and we should store the optional TimeValue as a field of DateValue without changing `value` at all.

In fact fact, now that I think about it, expressions like "in 2 days" are parsed as "$start_of(day) + 2 day" (today + 2 days) which is a ComputationValue. If we want to say "in 2 days at 5pm", we need to either:
1. attach the time to the first expression (so `new Date ( $start_of(day) , TIME_0 )  + 2 day )`) which would work for this specific case but also would not be very compositional
2. make a regular scalar operator that sets the time on a Date object: `setTime ( $start_of ( day ) ) + 2 day , TIME_0 )`

I think option 2 looks a bit cleaner. It would immediately support relative times and complex expressions. In the common case it has the same complexity in the neural model. It's a bit less intuitive though.

emilyjchang:
So it sounds like instead of having DateValue take a TimeValue, we should just use ComputationValue to figure out the date given, say a date and some relative/absolute times?

gcampax:
Yes I think that's what we will have to do in the fully general case.

gcampax (approved):
Yep! We can add a couple more tests - maybe one in test_syntax.tt and one in test_builtins.ts - but the implementation is all there.

PR #358, opened 2021-06-30
Add convenience APIs to access the current and next item in a dialogue

gcampax:
Current is the one that was just executed, next is the one immediately
after (ignoring proposed ones)

---

This APIs are currently part of the "ContextInfo" class which is a messy class internal
to the state machine in Genie. But they are generally useful and we want to expose
them to custom dialogues, so put them here to be cleaned and documented.

### DIFF
--- a/lib/ast/dialogues.ts
+++ b/lib/ast/dialogues.ts
@@ -333,12 +333,38 @@ export class DialogueHistoryItem extends AstNode {
  *
  */
 export class DialogueState extends Input {
+    /**
+     * The identifier of the current dialogue policy.
+     * This is a reverse-DNS name and acts as a namespace
+     * for the dialogue act.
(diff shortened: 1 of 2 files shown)


lib/ast/dialogues.ts
```diff
@@ -356,20 +382,85 @@ export class DialogueState extends Input {
...
+    get next() : DialogueHistoryItem|null {
+        return this._next;
+    }
+
+    /**
+     * The definition of the main Thingpedia function associated with the next
+     * statement, if any.
+     */
+    get nextFunction() : FunctionDef|null {
+        return this._current ? this._current.stmt.expression.schema : null;
```
rayslxu, line 446: this should be referring to `this._next`, right?

gcampax: Ugh yes

PR #361, opened 2021-07-12
Entity inheritance changes cherry-picked from CSQA branch

rayslxu:


lib/ast/class_def.ts
```diff
@@ -429,7 +429,7 @@ export class EntityDef extends Node {
      */
     constructor(location : SourceRange|null,
                 name : string,
-                _extends : string|null,
+                _extends : string[],
```
gcampax, line 432: I wonder if we can keep the existing constructor signature here, and automatically adjust the type when we set the property.

lib/type.ts
```diff
@@ -521,17 +521,37 @@ function arrayEquals(a : unknown[]|null, b : unknown[]|null) : boolean {
...
 }
 
-const DEFAULT_ENTITY_SUB_TYPE : Record<string, string> = {
-    'tt:picture': 'tt:url'
+const DEFAULT_ENTITY_SUB_TYPE : Record<string, string[]> = {
+    'tt:picture': ['tt:url']
 };
 
-function entitySubType(type : string, assignableTo : string, entitySubTypeMap : EntitySubTypeMap) : boolean {
+function _getAncestors(type : string, entitySubTypeMap : EntitySubTypeMap) : string[] {
```
gcampax, line 528: This doesn't seem very efficient, should we cache it somehow?

lib/type.ts
```diff
@@ -138,10 +138,10 @@ export default abstract class Type implements Hashable<Type> {
         return type;
     }
 
-    static isAssignable(type : Type, assignableTo : Type|string, typeScope : TypeScope = {}, entitySubTypeMap : EntitySubTypeMap = {}) : boolean {
+    static isAssignable(type : Type, assignableTo : Type|string, typeScope : TypeScope = {}, entitySubTypeMap : EntitySubTypeMap = {}, relax = false) : boolean {
```
gcampax, line 141: Ugh relax is back. What are the semantics of relax now?

gcampax (commented):
I think overall the design makes sense. Couple questions on the implementation.

lib/type.ts
rayslxu: > Ugh relax is back. What are the semantics of relax now?

when `relax` is true, `a` is assignable to `b` if they share common subtypes. I added `relax` to avoid breaking existing code just in case... 
i agree it's confusing... do you think we can make this the standard for everything?

lib/ast/class_def.ts
rayslxu: fixed in c7c5506

lib/type.ts
rayslxu: fixed in c33993f

lib/ast/class_def.ts
```diff
@@ -429,21 +429,23 @@ export class EntityDef extends Node {
...
         this.impl_annotations = annotations.impl || {};
     }
 
     toSource() : TokenStream {
-        if (this.extends) {
+        if (this.extends.length > 0) {
+            const _extends = this.extends.map((e) => e.includes(':') ? `^^${e}` : e);
             return List.concat('entity', ' ', this.name,
-                'extends', (this.extends.includes(':') ? '^^' + this.extends : this.extends), '\t+',
+                'extends', ..._extends.join(' , ').split(' '), '\t+',
```
gcampax, line 448: Uh? You should use List.join here, that's what it is for. Also, you still need the ^^ prefix if the entity name contains :

lib/typecheck.ts
```diff
@@ -368,6 +383,23 @@ export default class TypeChecker {
             throw new TypeError(`Invalid principal ${principal}, must be a contact or a group`);
     }
 
+    private _expandTypeScope(typeScope : TypeScope, key : string) : TypeScope[] {
+        if (!(key in typeScope))
+            return [typeScope];
+        if (!(typeScope[key] instanceof EntityType))
+            return [typeScope];
+        
+        const entityType = (typeScope[key] as EntityType).type;
```
gcampax, line 392: Do you need this type assertion? You did the instanceof check right before it, doesn't the compiler pick it up?
If it doesn't, maybe with a temporary variable?

lib/typecheck.ts
```diff
@@ -377,7 +409,16 @@ export default class TypeChecker {
             const typeScope : TypeScope = {};
             let good = true;
             for (let i = 0; i < argTypes.length; i++) {
-                if (!await this._isAssignable(argTypes[i], overload[i], typeScope)) {
+                const typeScopes = typeof overload[i] === 'string' ? 
+                    this._expandTypeScope(typeScope, overload[i] as string) : [typeScope];
```
gcampax, line 413: Same here, you have a typeof check

lib/typecheck.ts
```diff
@@ -368,6 +383,23 @@ export default class TypeChecker {
             throw new TypeError(`Invalid principal ${principal}, must be a contact or a group`);
     }
 
+    private _expandTypeScope(typeScope : TypeScope, key : string) : TypeScope[] {
```
gcampax, line 386: Can you add a comment explaining what this function does, and how it's used in _resolveOverload?

test/thingpedia.tt
```diff
@@ -5340,6 +5340,29 @@ class @com.test.entitysubtype2 {
   action foo(in req param : Entity(com.test.entitysubtype2:parent));
 }
 
+class @com.test.entitysubtype3 {
+  entity thing #[has_ner_support=false];
+
+  entity country #[has_ner_support=true];
```
gcampax, line 5346: `extends thing`?

gcampax (approved):
I think this looks good (minus a few comments). I am bit unclear about the logic used in the typechecking implementation, but it will be ok with a comment explaining it.

PR #362, opened 2021-07-12
Use stack instead of recursion for List iterator

rayslxu:
Otherwise, it causeses "RanegeError: Maximum call stack size exceeded"
when the list is long (e.g., 2000+)


gcampax (approved):
Yeah

PR #363, opened 2021-07-14
Add support for boolean questions

rayslxu:


gcampax (commented):
I don't quite understand the semantics here. Why are these compiled to filters and not something like a projection?
If the user asks "is mount everest taller than 9000m", you want to answer "no, mount everest is 8,849 meters". How do you that with the current compilation method?

rayslxu:
> I don't quite understand the semantics here. Why are these compiled to filters and not something like a projection?
> If the user asks "is mount everest taller than 9000m", you want to answer "no, mount everest is 8,849 meters". How do you that with the current compilation method?

Hmm, I forgot that we need provide result info even if the answer is no. 
What would you suggest to return? Maybe a `__result` field in addition to the normal return result? I'm not sure if I should do a projection on the compared field(s).

gcampax:
Yeah I'm thinking a boolean question should have a new field "answer" or similar that gets added, either to all fields in the normal return value, or just to the fields referenced in the boolean expression.

gcampax:
And we adjust the FunctionDef accordingly because otherwise Genie result handling code will be very confused.

rayslxu:
> And we adjust the FunctionDef accordingly because otherwise Genie result handling code will be very confused.

How about now?

lib/compiler/ops.ts
```diff
@@ -54,6 +54,16 @@ export class Compute extends PointWiseOp {
         return `PointWiseOp.Compute(${this.expression} as ${this.alias})`;
     }
 }
+
+export class Verification extends PointWiseOp {
```
gcampax, line 58: Why is this called "Verification"?

lib/compiler/ops-to-jsir.ts
```diff
@@ -825,6 +825,20 @@ export default class OpCompiler {
...
+    private _compileVerification(verification : PointWiseOp.Verification) {
+        const answer = this._compileFilter(compileBooleanExpressionToOp(verification.booleanExpression), this._currentScope);
+        this._irBuilder.add(new JSIr.SetKey(getRegister('$output', this._currentScope), '__verification_answer', answer));
+
+        this._currentScope.set('__verification_answer', {
+            type: 'scalar',
+            register: answer,
+            tt_type: Type.Boolean,
+            direction: 'output',
+            isInVarScopeNames: true
```
gcampax, line 837: Can you monitor() a boolean question? What are semantics?
(`isInVarScopeNames` controls whether a field affects monitors or not)

lib/new-syntax/parser.lr
```diff
@@ -887,6 +887,10 @@ chain_expression_list : Ast.Expression[] = {
 projection_expression : Ast.Expression = {
     filter_expression;
 
+    '[' boolean_expression: or_filter ']' 'of' expr:filter_expression => {
```
gcampax, line 890: Btw, do we need to support multiple boolean questions at once?

lib/compiler/ops.ts
rayslxu: it's verifying if a condition is true or not for a table? I just thought `Boolean` or `BooleanExpression` would be a weird here

lib/compiler/ops-to-jsir.ts
rayslxu: I misunderstood the meaning of this field then, why the name? also I don't think there is a comment about this confusing name...

lib/new-syntax/parser.lr
rayslxu: hmm, then we will need to somehow name the boolean questions since the order in natural language may not be the same as the order in code. 
we are going to have support for multiple queries, I'm tempted to keep this simple and use separate queries when we really need to...

lib/compiler/ops.ts
gcampax: Maybe BooleanComputation or similar? That's what it is anyway. Verification is weird, it sounds like you're enforcing some assertion.

lib/new-syntax/parser.lr
gcampax: Yeah I agree that if we want compound queries at some point we might as well handle them as two separate queries with a good old semicolon.

lib/compiler/ops-to-jsir.ts
gcampax: Yeah I think the name is terrible and it's partly historical, and I don't really remember the rationale for the name originally. I think originally we had a thing called "varScope" which was a record with all the variables in scope, and "varScopeNames" which was the list of names, used for monitoring. Then we got a boolean indicating whether something was in "varScopeNames", and eventually "varScopeNames" went away but the bad boolean remained.

I think the boolean properly means "is this an output parameter of the current ThingTalk expression" - as opposed to some special compiler-internal variable like $output, a let-assignment, or the input parameter of an enclosing procedure. As it happens, that is only really useful for monitoring though.

rayslxu:
@gcampax  I have renamed `PointWiseOp.Verification` to `PointWiseOp.BooleanCompute`, to be inconsistent with `PointWiseOp.Compute`. 
the special field is renamed to `__answer` instead of `__verification_answer`, if we need different types of answers in the future, we can change that, but let's keep it clean for now.

lib/new-syntax/parser.lr
```diff
@@ -922,9 +926,9 @@ projection_nonstar_param_list : Array<[Ast.Value, string|null]> = {
 }
 
 projection_param : [Ast.Value, string|null] = {
-    v:value => [v, null];
+    v:filter_lhs_filter_value => [v, null];
 
-    v:value 'as' alias:variable_name => [v, alias];
+    v:filter_lhs_filter_value 'as' alias:variable_name => [v, alias];
```
gcampax, line 931: This change means that `[true] of ...` will be parsed as a BooleanQuestionExpression instead of a ProjectionExpression. I think that's reasonable. Maybe a comment? This part of the grammar is quite tricky...

Btw, do we have an optimize() step to optimize the filter inside BooleanQuestionExpression? And what happens if the expression is statically true or statically false? Do we remove the boolean question or always answer yes/no?

gcampax (approved):
This looks good! I think we need more optimize code to bring BQEs to the top like projection expressions, but maybe it's not urgent.

PR #365, opened 2021-07-15
Unconditionally emit the output of actions

gcampax:
See the second commit for details. The first commit is prep work so the second one is not too tedious.


rayslxu:
how does the `updateMode` work? it seems only out the existing test cases out without modifying anything?

gcampax:
> how does the `updateMode` work? it seems only output the existing test cases without modifying anything?

test() modifies the TEST_CASE array in place with the generated output

rayslxu (approved):
looks good

PR #369, opened 2021-07-31
Fix default minimal projection if there is an ID input parameter

gcampax:
The minimal projection should not include "id" if "id" is not an output.

---

This currently trips the Genie basic tests in https://github.com/stanford-oval/genie-toolkit/pull/729 because basic tests fetch the snapshot from Thingpedia. Thingpedia snapshots don't include `#[minimal_projection]` so we get the default, and we have a function that has an `id` input parameter.


rayslxu (approved):
oops

PR #370, opened 2021-07-31
Allow arbitrary annotations on dialogues and dialogue history items

gcampax:
These can be useful to attach metadata associated with the execution
of certain ThingTalk statements, for example detailed error messages
and stack traces.

Annotations are dropped when the dialogue state is prepared for
the neural model so they don't affect parsing.

---

The goal of this PR is to get detailed error messages and stack traces in the conversation logs, so we can debug JS errors.
Given conversation logs are ThingTalk-based, it's convenient to add those as annotations.

The implementation is a bit messy because annotations are already attached to dialogue history items with special purposes, and we don't want to change the AST or the syntax of those.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ all = \
 	lib/grammar.js \
 	test/test_sr_parser_generator.ts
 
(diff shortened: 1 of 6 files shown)


rayslxu (approved):
lgtm

Issue #364, opened 2021-07-15
DialogueHistoryResultItem.clone is not a deep-clone


gcampax:
All Ast.Node.clone methods should be deep-clones and clone all descendant Ast nodes. DialogueHistoryResultItem instead shares the Ast Value which is a bug.

PR #380, opened 2021-08-20
Bring back join

rayslxu:
The new `join` returns a pair of results for each side of the join. The returned result is flattened, where the fields from the left-hand side are prefixed with `first.` and the fields from the right-hand side are prefixed with `second.` 
The syntax allows projection/filter over joins, however, in practice, they should all be pushed down to either the left-hand side or the right-hand side. This is not currently enforced. Not sure if we want to enforce it in typechecking or optimization, or simply leave it to genie.

### DIFF
--- a/lib/ast/expression.ts
+++ b/lib/ast/expression.ts
@@ -27,7 +27,7 @@ import {
     DeviceSelector,
     InputParam,
 } from './invocation';
-import { BooleanExpression, TrueBooleanExpression } from './boolean_expression';
+import { BooleanExpression } from './boolean_expression';
 import * as legacy from './legacy';
 import {
     Value,
@@ -1196,18 +1196,15 @@ export class ChainExpression extends Expression {
 export class JoinExpression extends Expression {
     lhs : Expression;
     rhs : Expression;
-    condition : BooleanExpression;
 
     constructor(location : SourceRange|null,
                 left : Expression,
                 right : Expression,
-                condition : BooleanExpression, 
                 schema : FunctionDef|null) {
         super(location, schema);
 
         this.lhs = left;
         this.rhs = right;
-        this.condition = condition;
     }
 
     get priority() : SyntaxPriority {
@@ -1215,26 +1212,16 @@ export class JoinExpression extends Expression {
     }
 
     toSource() : TokenStream {
-        if (this.condition instanceof TrueBooleanExpression) {
-            return List.concat(
-                addParenthesis(this.priority, this.lhs.priority, this.lhs.toSource()),
-                'join',
-                addParenthesis(this.priority, this.rhs.priority, this.rhs.toSource())
-            );
-        } else {
-            return List.concat(
-                addParenthesis(this.priority, this.lhs.priority, this.lhs.toSource()),
-                'join',
-                addParenthesis(this.priority, this.rhs.priority, this.rhs.toSource()),
-                'on',
-                addParenthesis(this.priority, this.condition.priority, this.condition.toSource())
-            );
-        }
+        return List.concat(
+            addParenthesis(this.priority, this.lhs.priority, this.lhs.toSource()),
+            'join',
+            addParenthesis(this.priority, this.rhs.priority, this.rhs.toSource())
+        );
     }
 
     equals(other : Expression) : boolean {
         return other instanceof JoinExpression &&
-            this.lhs.equals(other.lhs) && this.rhs.equals(other.rhs) && this.condition.equals(other.condition);
+            this.lhs.equals(other.lhs) && this.rhs.equals(other.rhs);
     }
 
     toLegacy(into_params : InputParam[] = [], scope_params : string[] = []) : legacy.Table {
@@ -1246,7 +1233,6 @@ export class JoinExpression extends Expression {
         if (visitor.visitJoinExpression(this)) {
             this.lhs.visit(visitor);
             this.rhs.visit(visitor);
-            this.condition.visit(visitor);
         }
         visitor.exit(this);
     }
@@ -1256,7 +1242,6 @@ export class JoinExpression extends Expression {
             this.location,
             this.lhs.clone(),
             this.rhs.clone(),
-            this.condition.clone(),
             this.schema ? this.schema.clone() : null
         );
     }
--- a/lib/ast/expression.ts
+++ b/lib/ast/expression.ts
@@ -27,7 +27,7 @@ import {
     DeviceSelector,
     InputParam,
 } from './invocation';
-import { BooleanExpression } from './boolean_expression';
+import { BooleanExpression, TrueBooleanExpression } from './boolean_expression';
 import * as legacy from './legacy';
 import {
     Value,
@@ -1188,6 +1188,92 @@ export class ChainExpression extends Expression {
             [, scope] = yield* expr.iterateSlots2(scope);
             Object.assign(newScope, scope);
         }
+        return [null, newScope];
+    }
+}
+
+
+export class JoinExpression extends Expression {
+    lhs : Expression;
+    rhs : Expression;
+    condition : BooleanExpression;
+
+    constructor(location : SourceRange|null,
+                left : Expression,
+                right : Expression,
+                condition : BooleanExpression, 
+                schema : FunctionDef|null) {
+        super(location, schema);
+
+        this.lhs = left;
+        this.rhs = right;
+        this.condition = condition;
+    }
+
+    get priority() : SyntaxPriority {
+        return SyntaxPriority.Join;
+    }
+
+    toSource() : TokenStream {
+        if (this.condition instanceof TrueBooleanExpression) {
+            return List.concat(
+                addParenthesis(this.priority, this.lhs.priority, this.lhs.toSource()),
+                'join',
+                addParenthesis(this.priority, this.rhs.priority, this.rhs.toSource())
+            );
+        } else {
+            return List.concat(
+                addParenthesis(this.priority, this.lhs.priority, this.lhs.toSource()),
+                'join',
+                addParenthesis(this.priority, this.rhs.priority, this.rhs.toSource()),
+                'on',
+                addParenthesis(this.priority, this.condition.priority, this.condition.toSource())
+            );
+        }
+    }
+
+    equals(other : Expression) : boolean {
+        return other instanceof JoinExpression &&
+            this.lhs.equals(other.lhs) && this.rhs.equals(other.rhs) && this.condition.equals(other.condition);
+    }
+
+    toLegacy(into_params : InputParam[] = [], scope_params : string[] = []) : legacy.Table {
+        throw new UnserializableError('Join expression');
+    }
+
+    visit(visitor : NodeVisitor) : void {
+        visitor.enter(this);
+        if (visitor.visitJoinExpression(this)) {
+            this.lhs.visit(visitor);
+            this.rhs.visit(visitor);
+            this.condition.visit(visitor);
+        }
+        visitor.exit(this);
+    }
+
+    clone() : JoinExpression {
+        return new JoinExpression(
+            this.location,
+            this.lhs.clone(),
+            this.rhs.clone(),
+            this.condition.clone(),
+            this.schema ? this.schema.clone() : null
+        );
+    }
+
+    *iterateSlots(scope : ScopeMap) : Generator<OldSlot, [InvocationLike|null, ScopeMap]> {
+        const [, leftScope] = yield* this.lhs.iterateSlots(scope);
+        const [, rightScope] = yield* this.rhs.iterateSlots(scope);
+        const newScope : ScopeMap = {};
+        Object.assign(newScope, leftScope, rightScope);
+        return [null, newScope];
+    }
+
+    *iterateSlots2(scope : ScopeMap) : Generator<DeviceSelector|AbstractSlot, [InvocationLike|null, ScopeMap]> {
+        const [, leftScope] = yield* this.lhs.iterateSlots2(scope);
+        const [, rightScope] = yield* this.rhs.iterateSlots2(scope);
+        const newScope : ScopeMap = {};
+        Object.assign(newScope, leftScope, rightScope);
         return [null, newScope];
     }
 }
--- a/lib/compiler/ast-to-ops.ts
+++ b/lib/compiler/ast-to-ops.ts
@@ -652,14 +652,13 @@ function compileTableToOps(table : Ast.Expression,
     } else if (table instanceof Ast.JoinExpression) {
         const lhsop = compileTableToOps(table.lhs, restrictHintsForJoin(hints, table.lhs.schema!));
         const rhsop = compileTableToOps(table.rhs, restrictHintsForJoin(hints, table.rhs.schema!));
-        const condition = table.condition ? compileBooleanExpressionToOp(table.condition) : null;
         let device : Ast.DeviceSelector|null = null;
         let handle_thingtalk = false;
         if (lhsop.device && rhsop.device) {
             device = sameDevice(lhsop.device, rhsop.device) ? lhsop.device : null;
             handle_thingtalk = sameDevice(lhsop.device, rhsop.device) ? lhsop.handle_thingtalk && rhsop.handle_thingtalk : false;
         }
-        return new TableOp.Join(lhsop, rhsop, condition, device, handle_thingtalk, table);
+        return new TableOp.Join(lhsop, rhsop, device, handle_thingtalk, table);
     } else {
         throw new TypeError(table.constructor.name);
     }
--- a/lib/compiler/ast-to-ops.ts
+++ b/lib/compiler/ast-to-ops.ts
@@ -649,6 +649,17 @@ function compileTableToOps(table : Ast.Expression,
             compiled.handle_thingtalk, 
             table
         );
+    } else if (table instanceof Ast.JoinExpression) {
+        const lhsop = compileTableToOps(table.lhs, restrictHintsForJoin(hints, table.lhs.schema!));
+        const rhsop = compileTableToOps(table.rhs, restrictHintsForJoin(hints, table.rhs.schema!));
+        const condition = table.condition ? compileBooleanExpressionToOp(table.condition) : null;
+        let device : Ast.DeviceSelector|null = null;
+        let handle_thingtalk = false;
+        if (lhsop.device && rhsop.device) {
+            device = sameDevice(lhsop.device, rhsop.device) ? lhsop.device : null;
+            handle_thingtalk = sameDevice(lhsop.device, rhsop.device) ? lhsop.handle_thingtalk && rhsop.handle_thingtalk : false;
+        }
+        return new TableOp.Join(lhsop, rhsop, condition, device, handle_thingtalk, table);
     } else {
         throw new TypeError(table.constructor.name);
     }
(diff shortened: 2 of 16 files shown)


lib/new-syntax/parser.lr
```diff
@@ -920,11 +920,27 @@ toplevel_chain_expression : Ast.ChainExpression = {
...
+}
+
+join_expression : Ast.Expression = {
+    lhs:projection_expression 'join' rhs:projection_expression => {
+        return new Ast.JoinExpression($.location, lhs, rhs, null, null);
+    };
+
+    lhs:projection_expression 'join' rhs:projection_expression 'on' condition:or_filter => {
+        return new Ast.JoinExpression($.location, lhs, rhs, condition, null);
+    };
```
gcampax, line 943: The way you wrote the grammar, `A join B join C` is not allowed, but the way you wrote toSource(), `(A join B) join C` will turn into `A join B join C` without parenthesis.

I think the right way to write the grammar is:
```
chain_expression_list = {
   join_expression;
   chain_expression_list => join_expression;
}
join_expression = {
   projection_expression;
   join_expression 'join' projection_expression;
   join_expression 'join' projection_expression 'on' or_filter;
}
```

lib/new-syntax/pretty.ts
```diff
@@ -274,5 +276,5 @@ export function prettyprint(tokens : TokenStream) : string {
         }
     }
 
-    return buffer.trim();
+    return buffer.replace(/ +/g, ' ').trim();
```
gcampax, line 279: In what case do we generate extra spaces?

lib/optimize.ts
```diff
@@ -523,6 +523,13 @@ function optimizeExpression(expression : Ast.Expression, allow_projection=true)
         return expression;
     }
 
+    if (expression instanceof Ast.JoinExpression) {
+        const lhs = optimizeExpression(expression.lhs);
+        const rhs = optimizeExpression(expression.rhs);
+        const condition = expression.condition ? optimizeFilter(expression.condition) : null;
```
gcampax, line 529: The order of the operands matter for the join condition. Are you sure you can optimize blindly?
Maybe add a test case?

lib/ast/expression.ts
```diff
@@ -1191,3 +1191,88 @@ export class ChainExpression extends Expression {
...
+
+export class JoinExpression extends Expression {
+    lhs : Expression;
+    rhs : Expression;
+    condition : BooleanExpression|null;
+
+    constructor(location : SourceRange|null,
+                left : Expression,
+                right : Expression,
+                condition : BooleanExpression|null,
```
gcampax, line 1204: Should condition be Ast.BooleanExpression.True instead of null?

lib/ast/expression.ts
```diff
@@ -1191,3 +1191,88 @@ export class ChainExpression extends Expression {
...
+                addParenthesis(this.priority, this.lhs.priority, this.lhs.toSource()),
+                'join',
+                addParenthesis(this.priority, this.rhs.priority, this.rhs.toSource())
+            );
+        }
+    }
+
+    equals(other : Expression) : boolean {
+        return other instanceof JoinExpression &&
+            this.lhs === other.lhs && this.rhs === other.rhs && this.condition === other.condition;
```
gcampax, line 1237: All the === needs to be .equals() instead.

lib/ast/expression.ts
```diff
@@ -1191,3 +1191,88 @@ export class ChainExpression extends Expression {
...
+    }
+
+    toLegacy(into_params : InputParam[] = [], scope_params : string[] = []) : legacy.Table {
+        throw new Error('Not supported by legacy ThingTalk');
```
gcampax, line 1241: We have a dedicated error class for this.

lib/ast/expression.ts
```diff
@@ -1191,3 +1191,88 @@ export class ChainExpression extends Expression {
...
+    }
+
+    visit(visitor : NodeVisitor) : void {
+        visitor.enter(this);
```
gcampax, line 1245: The visitor pattern is not correct: you need:
```
if (visitor.visitJoinExpression(this)) {
   this.lhs.visit(visitor);
   ...
}
```

lib/ast/expression.ts
```diff
@@ -1191,3 +1191,88 @@ export class ChainExpression extends Expression {
...
+            this.condition.visit(visitor);
+        visitor.exit(this);
+    }
+
+    clone() : JoinExpression {
+        return new JoinExpression(
+            this.location,
+            this.lhs,
+            this.rhs,
+            this.condition,
```
gcampax, line 1258: clone() must be a deep-clone (see also #364 )

lib/compiler/ast-to-ops.ts
```diff
@@ -649,6 +649,17 @@ function compileTableToOps(table : Ast.Expression,
...
+        const lhsop = compileTableToOps(table.lhs, restrictHintsForJoin(hints, table.lhs.schema!));
+        const rhsop = compileTableToOps(table.rhs, restrictHintsForJoin(hints, table.rhs.schema!));
+        const condition = table.condition ? compileBooleanExpressionToOp(table.condition) : null;
+        let device : Ast.DeviceSelector|null = null;
+        let handle_thingtalk = false;
+        if (lhsop.device && rhsop.device) {
+            device = sameDevice(lhsop.device, rhsop.device) ? lhsop.device : null;
+            handle_thingtalk = sameDevice(lhsop.device, rhsop.device) ? lhsop.handle_thingtalk && rhsop.handle_thingtalk : false;
+        }
+        return new TableOp.Join(lhsop, rhsop, condition, device, handle_thingtalk, table);
```
gcampax, line 662: Why do we need a new TableOp? NestedLoopJoin or CrossJoin should have the right semantics, because they were written at the time we had joins in ThingTalk 1.0

In fact, as a far as the compiler is concerned, a join is a CrossJoin op wrapped with a FilterOp, isn't it?

lib/runtime/stream_table_ops.ts
```diff
@@ -167,6 +167,31 @@ class DelayedIterator<T> implements AsyncIterator<T> {
     }
 }
 
+export function tableJoin<T>(lhs : Stream<T>, rhs : Stream<T>) : AsyncIterator<ResultT<T>, void> {
```
gcampax, line 170: The type signature is not correct, it should be:
```
export function tableJoin<T1, T2>(lhs : Stream<T1>, rhs : Stream<T2>) : AsyncIterator<ResultT<T1 & T2>, void> {
```
given that it takes two streams and computes a stream with both properties.

lib/typecheck.ts
```diff
@@ -679,6 +679,61 @@ export default class TypeChecker {
         return Promise.resolve();
     }
 
+    private async _typeCheckJoin(ast : Ast.JoinExpression, scope : Scope) {
+        assert(ast.lhs.schema && ast.rhs.schema);
+        const name = `join(${ast.lhs.schema.name},${ast.rhs.schema.name})`;
+        if (!ast.lhs.schema.is_list || !ast.rhs.schema.is_list)
+            throw new TypeError('Both sides of a join must be a list query');
```
gcampax, line 686: Why this restriction?

lib/typecheck.ts
```diff
@@ -679,6 +679,61 @@ export default class TypeChecker {
...
+        const classDef = null; // FIXME: what should this be for a join expression?
```
gcampax, line 687: It definitely should be null (this will cause issues in the dialogue code but we'll handle those)

lib/typecheck.ts
```diff
@@ -679,6 +679,61 @@ export default class TypeChecker {
         return Promise.resolve();
     }
 
+    private async _typeCheckJoin(ast : Ast.JoinExpression, scope : Scope) {
+        assert(ast.lhs.schema && ast.rhs.schema);
+        const name = `join(${ast.lhs.schema.name},${ast.rhs.schema.name})`;
```
gcampax, line 684:
```suggestion
        const name = `join(${ast.lhs.schema.qualifiedName},${ast.rhs.schema.qualifiedName})`;
```

lib/typecheck.ts
```diff
@@ -679,6 +679,61 @@ export default class TypeChecker {
...
+            if (!schema.hasArgument(ast.value.name))
+                throw new TypeError(`Field ${ast.value.name.slice('second.'.length)} is not available on the left-hand side of the join`);
+            const type_lhs = schema.getArgType(ast.name);
+            const type_rhs = schema.getArgType(ast.value.name);
+            ast.overload = await this._resolveFilterOverload(type_lhs!, ast.operator, type_rhs!);
+        } else if (ast instanceof Ast.NotBooleanExpression) {
+            this._typeCheckJoinCondition(ast.expr, schema);
+        } else if (ast instanceof Ast.AndBooleanExpression || ast instanceof Ast.OrBooleanExpression) {
+            ast.operands.forEach((o) => this._typeCheckJoinCondition(o, schema));
+        } else {
```
gcampax, line 732: True and False?

lib/typecheck.ts
```diff
@@ -679,6 +679,61 @@ export default class TypeChecker {
...
+        ast.schema = new Ast.FunctionDef(null, 'query', classDef, name, [], qualifiers, args);
+        if (ast.condition) 
+            await this._typeCheckJoinCondition(ast.condition, ast.schema);
+        return Promise.resolve();
+    }
+
+    private async _typeCheckJoinCondition(ast : Ast.BooleanExpression, schema : Ast.FunctionDef) {
+        if (ast instanceof Ast.AtomBooleanExpression) {
+            if (!ast.name.startsWith('first.'))
+                throw new TypeError('Fields on the left-hand side in a join condition needs a `first.` prefix');
```
gcampax, line 716: Why do you need the `first.` prefix? I thought the point of the special join syntax was to avoid `first.` and `second.` in the common cases by relying on the operand order...

test/test_compiler.txt
```diff
@@ -11508,3 +11508,146 @@ ontimer(date=[set_time(new Date(2021, 0, 1), new Time(10, 30, 0))]) => @org.thin
...
+            _t_27 = _t_22.popularity;
+            __emit(_t_21, _t_22);
+            _iter_tmp = await _t_19.next();
+          }
+        }
+      } catch(_exc_) {
+        __env.reportError("Failed to invoke query", _exc_);
+      }
+    }
+    _t_28 = __builtin.tableJoin(_t_0, _t_16);
```
gcampax, line 11617: You'll want to add a test for this builtin function in test_builtin.js, and also perhaps a runtime test for this program in test_runtime.js

test/test_compiler.txt
```diff
@@ -11508,3 +11508,146 @@ ontimer(date=[set_time(new Date(2021, 0, 1), new Time(10, 30, 0))]) => @org.thin
...
+        _t_36 = _t_31.genres;
+        _t_37 = _t_31.release_date;
+        _t_38 = _t_31.popularity;
+        _t_39 = _t_31.energy;
+        _t_40 = _t_31.danceability;
+        _t_41 = _t_31.__response;
+        _t_42 = _t_31.id;
+        _t_43 = _t_31.artists;
+        _t_44 = _t_31.release_date;
+        _t_45 = _t_31.popularity;
```
gcampax, line 11637: This code doesn't look correct. It should read `first.id`, `first.artists`, ... `second.id`, `second.artists`, ...

test/test_syntax.tt
```diff
@@ -2015,3 +2015,64 @@ $dialogue @org.thingpedia.dialogue.transaction;
...
+
+// ** expect TypeError **
+// Join with invalid condition 2
+@org.wikidata.city() join @org.wikidata.city() on first.id =~ "xyz";
+
+====
+
+// ** expect TypeError **
+// wrong projection over join
+[id] of (@org.wikidata.city() join @org.wikidata.city() on in_array(first.id, second.twinned_administrative_body));
```
gcampax, line 2078: If this projection is disallowed, why does __builtin.tableJoin include all the unprefixed parameters?
Having extra fields in the output object that are not listed in the FunctionDef will confuse the executor code in Genie.

lib/new-syntax/pretty.ts
rayslxu: `enum on`

lib/typecheck.ts
rayslxu: I mentioned this in the commit message. The observation from examples is that it's actually less readable and more confusing without them. Another minor issue is that it's a bit weird that the fields do not match the schema when compiling the program.

lib/new-syntax/pretty.ts
rayslxu: but apparently this is bad... we have indents, maybe `(?<!\s) +` would work?

rayslxu: but apparently this won't work as we have another space after `on` and there are also indents. 
I updated this to check if the buffer ends with `enum`. Is there any other case when a keyword is not used as a keyword?

lib/optimize.ts
rayslxu: because `first` comes before `second` so it should be fine (assuming we are keeping the prefix in the condition, if that is removed, yeah, then I need some special handling.

lib/compiler/ast-to-ops.ts
rayslxu: The semantic are different. Cross join will merge the scope (conflict fields on the lhs will be dropped), while `join` will keep both with prefix.

lib/typecheck.ts
rayslxu: Technically, it's impossible if all lhs is a field from one table, and rhs is a field from another table...

rayslxu: But given that you suggest to use True/False boolean expression instead of `null` for the condition, it makes sense to add them here

lib/new-syntax/parser.lr
rayslxu: I actually changed the addParentheses function, so it will keep the parentheses in `(A join B) join C`. but your design is better

rayslxu: hmm, the grammar gives me the following error

lib/ast/expression.ts
rayslxu: fixed in 61a6e46.

lib/ast/expression.ts
rayslxu: fixed in 61a6e46.

lib/ast/expression.ts
rayslxu: which one is it? The ones in `errors.ts` don't seem to be relevant?

lib/ast/expression.ts
rayslxu: fixed in 61a6e46.

lib/ast/expression.ts
rayslxu: fixed in 61a6e46.

lib/typecheck.ts
rayslxu: fixed in 61a6e46.

lib/typecheck.ts
rayslxu: removed in 2f7b9f5f

lib/typecheck.ts
rayslxu: comment removed in 2f7b9f5f

lib/typecheck.ts
rayslxu: fixed in 2f7b9f5f

test/test_compiler.txt
rayslxu: fixed in fff7277

test/test_syntax.tt
rayslxu: fixed in ebb30eb

test/test_compiler.txt
rayslxu: builtin test added in ebb30eb.

lib/runtime/stream_table_ops.ts
rayslxu: fixed in 2e7345d

lib/new-syntax/parser.lr
gcampax: The right-hand-side of 'join' should be a "projection_expression" not a "join_expression". `join` is left-associative in the grammar.

lib/new-syntax/pretty.ts
gcampax: I think units have special keyword behavior but those are handled differently anyway, so yeah that should be it?

lib/ast/expression.ts
gcampax: I think we used UnserializableError in other situations? toLegacy is only used when serializing to legacy syntax anyway so the error would make sense.

lib/compiler/ast-to-ops.ts
gcampax: I think we need to rename this stuff, because it's confusing. Also, CrossJoin for => doesn't really make sense, we should make => always use the nested-loop algorithm.

lib/typecheck.ts
gcampax: But then, what's the difference between:
```
@q1() join @q2() on first.foo = second.bar
```
and
```
(@q1() join @q2()) filter first.foo = second.bar
```
?

The special join syntax is unnecessary.

test/test_compiler.txt
```diff
@@ -11508,3 +11508,146 @@ ontimer(date=[set_time(new Date(2021, 0, 1), new Time(10, 30, 0))]) => @org.thin
...
+    }
+    _t_28 = __builtin.tableJoin(_t_0, _t_16);
+    {
+      let _iter_tmp = await _t_28.next();
+      while (!_iter_tmp.done) {
+        _t_29 = _iter_tmp.value;
+        _t_30 = _t_29[0];
+        _t_31 = _t_29[1];
+        _t_32 = _t_31.first.__response;
+        _t_33 = _t_31.first.id;
```
gcampax, line 11625: Wait no!
This needs to be `_t_31["first.id"]`, because there is no "first" object, it's a flat set of prefixed properties.
Which, btw, will cause issues with the dialogue result syntax I suspect, because you have identifiers with dots in them. Can you add a test for that one too?

lib/typecheck.ts
rayslxu: good point... I think at this point I should just drop that special syntax

lib/compiler/ast-to-ops.ts
gcampax: Sure, separate PR is fine. But we really need to fix this or we'll forget what's going on and nobody else will be able to touch this code.

lib/new-syntax/keywords.ts
```diff
@@ -49,6 +49,7 @@ export const KEYWORDS = new Set<string>([
     'opt',
     'req',
     'sort',
+    'on',
```
gcampax, line 52: 'on' is not a keyword there is no special join filter syntax, right?

(maybe we should keep it as reserved word for the future anyway though)

gcampax (commented):
Do you mind rebasing this so the "address PR comments" fixes are in the appropriate original commits? Thanks!

rayslxu:
> 
> 
> Do you mind rebasing this so the "address PR comments" fixes are in the appropriate original commits? Thanks!

ugh...ok.. I need to split that commit...

lib/new-syntax/keywords.ts
rayslxu: yeah, that's what I thought, it's probably good to keep it as keyword

gcampax (approved):

PR #397, opened 2021-10-29
Misc fixes

gcampax:
Two major ones:
- making =~ pluggable by adding a new method to the Entity class
- fixing the spam of notifications when a new monitor is added


rayslxu (approved):
Looks good. But what's the motivation to introduce the StringLike type?

gcampax:
It's to allow overriding softmatch of genres in Spotify, which are strings, not entities.The skill can return an array of StringLike objects with a custom softmatch, instead of an array of strings.

PR #399, opened 2021-11-20
Update partial date support

emilyjchang:
- Add future partial date so that in order to get the next instance of a date (e.g. asking for "the 9th" when it's October 10th should return November 9th).
- Support for resolving a partial date when given only time

### DIFF
--- a/lib/utils/date_utils.ts
+++ b/lib/utils/date_utils.ts
@@ -71,11 +71,67 @@ function createEdgeDate(edge : ('start_of'|'end_of'), unit : string, timezone : 
     return new Date(date.epochMilliseconds);
 }
 
-function createDatePiece(year : number|null, month : number|null, day : number|null, time : AbsoluteTime|null, timezone : string) : Date {
+function getTimeMs(hour : number, minute : number, second : number) {
+    const hourMs = hour * 1000 * 60 * 60;
+    const minMs = minute * 1000 * 60;
+    const secMs = second * 1000;
+    return hourMs + minMs + secMs;
+}
+
+function createDatePiece(year : number|null, month : number|null, day : number|null, time : AbsoluteTime|null, timezone: string, future ?: boolean) : Date {
+    let date = Temporal.Now.zonedDateTimeISO(timezone);
+    //Get the next instance of the supplied values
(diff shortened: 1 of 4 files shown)


lib/utils/date_utils.ts
```diff
@@ -71,11 +71,67 @@ function createEdgeDate(edge : ('start_of'|'end_of'), unit : string, timezone :
...
+            if (time !== null) {
+                const curTime = getTimeMs(date.hour, date.minute, date.second);
+                const timeMs = getTimeMs(time.hour, time.minute, time.second);
+                //if the desired time is in the past for this day, get tomorrow
+                if (curTime > timeMs) {
+                    date.withPlainDate({
+                        year: date.year,
+                        month: date.month,
+                        day: date.day + 1
+                    });
```
gcampax, line 96: This code is not correct to compute tomorrow, you need to use `.add({ days: 1 })` because the month/year might change too.

lib/utils/date_utils.ts
```diff
@@ -71,11 +71,67 @@ function createEdgeDate(edge : ('start_of'|'end_of'), unit : string, timezone :
...
+                }
+            }
+        } else if (month === null) {
+            //the desired day is in the past for this month, get the next month
+            if (date.day > day) { 
+                date.withPlainDate({
+                    year: date.year,
+                    month: date.month + 1,
+                    day
+                });
```
gcampax, line 106: Same here, the year might change.

lib/utils/date_utils.ts
```diff
@@ -71,11 +71,67 @@ function createEdgeDate(edge : ('start_of'|'end_of'), unit : string, timezone :
...
+                    date.withPlainDate({
+                        year: date.year,
+                        month: date.month,
+                        day: date.day + 1
+                    });
+                }
+            }
+        } else if (month === null) {
+            //the desired day is in the past for this month, get the next month
+            if (date.day > day) {
```
gcampax, line 101: What if the day is equal? Should we add a month too?

gcampax (commented):
Thanks for working on this. I think we need to add some test cases to check this logic.

